PR #95, opened 2024-07-31
testing changes

cobac:
First change is about making sure the tests are calling `dbt-bouncer` as any other user installing this package would. More info in the commit description.

About the second change, I'm not sure about the current approach. Currently, the github-specific functionality is part of the `dbt-bouncer` python package. I think ideally it'd be another package/script that is included in the action alongside the `dbt-bouncer` package. In any case this is an immediate fix so that tests can run locally without the `GITHUB_TOKEN` envvar


cobac:
the second commit is mehing with the coverage but hey

pgoslatara (approved):
> First change is about making sure the tests are calling dbt-bouncer as any other user installing this package would. More info in the commit description.

Thanks for the references, super interesting reads.

> About the second change, I'm not sure about the current approach. Currently, the github-specific functionality is part of the dbt-bouncer python package. I think ideally it'd be another package/script that is included in the action alongside the dbt-bouncer package. In any case this is an immediate fix so that tests can run locally without the GITHUB_TOKEN envvar

Agreed on this. Long term, my idea is to add a feature that saves the results to a file and have a dedicated script read this file and post the GitHub comment. This script would live outside the python package and would be called as a second step in `action.yml`. But this depends on what we put in `action.yml`, currently it's a Docker image, but we can decide on the best approach at a later stage.

PR #99, opened 2024-08-01
Adding check on bytes billed

pgoslatara:


src/dbt_bouncer/checks/checks.md
```diff
@@ -163,6 +163,28 @@ manifest_checks:
...
 ---
 
+# `check_run_results_max_gigabytes_billed`
+
+Each result can have a maximum number of gigabytes billed. Note that this only works for the `dbt-bigquery` adapter.
+
+**Argument(s)**:
+
+* `max_gigabytes_billed`: The maximum execution time (seconds) allowed for a node.
+
```
cobac, line 173: typo!

pgoslatara: Silly me, correction in #102. Yet another reason that documentation should be auto-generated from doc strings and field descriptions.

cobac: I started working on it yesterday, #96

pgoslatara: Awesome! I've no experience with automated documentation so interested to see if this can work well or if we should take a roll-our-own approach.

PR #107, opened 2024-08-05
Add check match code

pgoslatara:


src/dbt_bouncer/checks/checks.md
```diff
@@ -131,6 +131,29 @@ manifest_checks:
 
 ---
 
+# `check_model_code_does_not_contain_regexp_pattern`
+
+The raw code for a model must not match the specified regexp pattern.
+
+**Argument(s)**:
+
+* `include`: (Optional) Regex pattern to match the macro path. Only macro paths that match the pattern will be checked.
```
cobac, line 140: s/macro path/model path

but could also be useful to validate macros?

PR #200, opened 2024-08-30
executable via uv

cobac:
wow

(`--no-cache` is just to showcase a fresh installation)

image

just works 

I think we can remove PEX completely from the repo 👀


pgoslatara:
`uv` looks awesome! Let's keep pex for now as it's baked into the Docker image, maybe in future we can use `uv` in the image.

PR #229, opened 2024-09-09
tldr and faq before cli docs

cobac:
comments in commits, feel free to ignore or change :)


cobac:
Could be added / rewrite the getting started section instead, but I like TLDRs on the readme when browsing

pgoslatara:
Added 42e2755; small bit of formatting but also added a `pip install ...` command and a `run_results_checks` check that can be used with any dbt adapter.

pgoslatara (approved):

PR #390, opened 2025-07-02
Allowing description for every check

### LINKED ISSUE #389: [Feature] Add optional `description` field to all checks
### Describe the feature

A lot of the checks are very flexible, but sometimes when using non-trivial regexes it is hard for people that did not write the regex to understand what the check is actually checking.

My proposal is to add an optional `description` field to all checks. If that field is populated, if the check fails instead of returning the default assertion error it would return the custom description.

E.g., it'll allow 

> 
> | Check name                           | Severity   | Failure message                                                                                                 |
> |--------------------------------------|------------|-----------------------------------------------------------------------------------------------------------------|
> | check_model_names:4:model_name | error      | AssertionError: `model_name` does not match the supplied regex `^stg(_[a-z0-9]+)+_(_[a-z0-9]+)+$`. |
> 

to become 

> 
> | Check name                           | Severity   | Failure message                                                                                                 |
> |--------------------------------------|------------|-----------------------------------------------------------------------------------------------------------------|
> | check_model_names:4:model_name | error      | Staging models should be called `stg_source_name__table_name`. |

if the developer configures the test like:

```yaml
  - name: check_model_names
    include: ^models/staging
    description: Staging models should be called `stg_source_name__table_name`.
    model_name_pattern: ^stg(_[a-z0-9]+)+_(_[a-z0-9]+)+$

```

### Are you interested in contributing this feature?

yes

pgoslatara:
Resolves #389. One change from the request, when a check fails, the description _and_ the failure message are returned. This is because the failure message contains detailed information like the model/macro/column/etc. that caused the failure and makes it much easier to identify what needs to be fixed.

### DIFF
--- a/dbt-bouncer-example.yml
+++ b/dbt-bouncer-example.yml
@@ -122,6 +122,7 @@ manifest_checks:
     include: ^models/intermediate
     model_name_pattern: ^int_
   - name: check_model_names
+    description: Models in the staging layer should always start with "stg_". # Descriptions are optional for all checks.
     include: ^models/staging
     model_name_pattern: ^stg_
   - name: check_model_test_coverage
--- a/docs/config_file.md
+++ b/docs/config_file.md
@@ -40,6 +40,29 @@ For more example config files, see [here](https://github.com/godatadriven/dbt-bo
 
 ## Common arguments
 
+### Description
+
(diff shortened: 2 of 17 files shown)


pgoslatara:
Thanks @cobac! Available from [1.14.0](https://github.com/godatadriven/dbt-bouncer/releases/tag/v1.14.0).

PR #522, opened 2025-12-18
refactor: rework internal check loading

mattcristal:
## Issues

- Parameters like `import_check` cause the same file to be re-imported as a new module for every single `Check` class found in it.
- The `get_check_objects` function returns a list of Module objects masquerading as "objects containing the class", which `config_file_parser.py`then has to decode.

## What was done

Refactor  `get_check_objects`

- Iterate over check files.
- Import each file once as a module.
- Inspect the module to find all classes starting with "Check"
- Collect these classes directly into the return list.
- Remove the import_check helper and the sys._getframe hack.
- Ensure custom checks are also loaded efficiently (one module load per file).

## Tests

Locally, ran:

```shell
make test
```


mattcristal:
pre-commit is passing now

Screenshot 2025-12-19 at 08 54 28

mattcristal:
With `prek`

Screenshot 2025-12-19 at 09 09 49

PR #672, opened 2026-02-24
refactor: use ABCs in seed checks

mattcristal:
## What was done

The current approach is:

A big `BaseCheck` class is inherited by all the check classes per Model, Seeds etc.

Instead, we could make use of some Abstract Bases Classes (ABCs) that have some common characteristics that can be then inherited in the various check classes. 

This is a suggestion. 

The current implementation is just done for the Seeds to demonstrate what is possible.

https://github.com/godatadriven/dbt-bouncer/pull/668

☝🏼 The one above is another approach, but it seems maybe more complex.

## Testing

Successfully run `prek` and `make test`

### DIFF
--- a/src/dbt_bouncer/check_base.py
+++ b/src/dbt_bouncer/check_base.py
@@ -1,9 +1,16 @@
+import logging
+import re
+from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Any, ClassVar, Literal
 
-from pydantic import BaseModel, ConfigDict, Field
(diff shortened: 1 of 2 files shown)


pgoslatara:
> @pgoslatara what do you think?

I like this pattern. Unfortunately the codebase has had significant changes since this PR was created so I don't think it makes sense to continue with this PR. Instead I've created https://github.com/godatadriven/dbt-bouncer/pull/734, heavily inspired by this PR. Thanks!

PR #706, opened 2026-02-27
refactor: split check_models.py into 11 focused files

pgoslatara:
## Summary
- Splits the 2100-line `check_models.py` into 11 topic-focused modules (5 already existed from a prior commit, 6 new ones added here)
- All 36 `CheckModel*` classes are preserved verbatim
- Auto-discovery via `get_check_objects()` glob means no import changes needed elsewhere
- Deletes the original `check_models.py`
- Updates `test_models.py` to import from the new split modules

## Test plan
- [ ] All 36 `CheckModel*` classes discovered by `get_check_objects()`
- [ ] All unit tests pass (461 passed)
- [ ] Pre-commit hooks pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -165,6 +165,7 @@ benchmark_results.json
 changed_files.txt
 coverage.json
 dbt.duckdb
+docs/plans
 github-comment.md
 pytest-coverage.txt
 results.json


mattcristal:
@pgoslatara 

Why not create a sub-folder `models` under `manifest`?

And then additional sub-folders like `seeds`?

PR #846, opened 2026-04-14
feat: add pre commit for json schema check

mattcristal:
## What was done

- Add a pre-commit for JSON schemas.
- Makes the developer's life a bit easier, automating this check.
- A pre-commit already exists for running alphabetical check script file.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -68,7 +68,7 @@ repos:
       hooks:
           - id: schema-check
             entry: bash -c 'uv run python scripts/generate_schema.py && git diff --exit-code schema.json || { echo "schema.json is out of date. Run \"make generate-schema\" and commit the result."; exit 1; }'
-            files: ^src/dbt_bouncer/checks/|^src/dbt_bouncer/configuration_file/|^scripts/generate_schema\.py$
+            files: ^schema\.json$|^scripts/generate_schema\.py$|^src/dbt_bouncer/checks/|^src/dbt_bouncer/configuration_file/
             language: system
             name: schema-check
             pass_filenames: false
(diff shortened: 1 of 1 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Adds a pre-commit hook to ensure `schema.json` stays in sync with the schema generation script and check/config source files.**

**Type:** `chore`

### Summary
This PR adds a new local pre-commit hook (`schema-check`) that runs `scripts/generate_schema.py` and then verifies `schema.json` has no uncommitted diff. If the schema is stale, the hook fails with an actionable error message. The second commit fixed the `files` pattern to also trigger when `schema.json` itself is staged (not just the Python sources that generate it).

### File Changes
| `.pre-commit-config.yaml` | Modified | Added `schema-check` local hook that regenerates the JSON schema and fails if the result differs from the committed version |

### Review Findings
- 🟡 **Suggestion:** `.pre-commit-config.yaml:70` — The `entry` uses `bash -c '...'` with an inline script containing escaped quotes and chained operators. This works but is fragile to maintain. Consider extracting the logic into a small shell script (e.g. `scripts/check_schema.sh`) for readability and easier future edits.
- 🟡 **Suggestion:** `.pre-commit-config.yaml:71` — The `files` pattern triggers on any file under `src/dbt_bouncer/checks/` and `src/dbt_bouncer/configuration_file/`, which is correct for detecting changes that could affect the schema. However, if new Pydantic models that feed the schema are added in other directories (e.g. `src/dbt_bouncer/models/`), the hook won't trigger. Worth documenting the assumption or using `pass_filenames: false` with `always_run: true` (trading speed for safety) if schema drift has been a recurring issue.
- ✅ No issues with security (no secrets, no unsafe deserialization, the `bash -c` invocation only runs local project scripts)
- ✅ No issues with breaking changes (additive-only change to pre-commit config)
- ✅ No issues with style (hook keys are alphabetically ordered, consistent with adjacent hooks)

---

pgoslatara (approved):

PR #883, opened 2026-05-12
fix: include internal check files in registry cache fingerprint

pgoslatara:
The disk cache fingerprint for the check-name → module map only covered the package version, custom-checks dir, and entry-point names — it didn't track the packaged `checks/` tree itself. In editable installs the version stays at `"0.0.0"`, so adding a new internal check leaves a previously-written cache valid by fingerprint but missing the new name. `get_check_objects_for_names()` then hits the `missing` branch and falls back to the full `get_check_objects()` scan, defeating the whole point of the targeted-import optimisation.

I noticed this while profiling the CLI. With a stale cache the example config takes ~409 ms inside `run_bouncer()`; with a freshly-built one (88 entries vs. my 85-entry stale file, 3 names missing) the same call drops to ~201 ms — a ~50% reduction in real work for users running any config that references a recently-added check.

The fix adds the packaged `checks/` tree (relative path + mtime per file) into the fingerprint so any check addition or edit invalidates the cache. The stat walk is ~0.6 ms for 37 files, well below the noise floor. As a bonus I also prune sibling `check_registry_*.json` files when writing a new one — my dev cache had 75 accumulated stale fingerprints — so the cache dir stays at exactly one entry per active interpreter.

Two new unit tests cover (1) that touching an internal check file changes the fingerprint and (2) that `_prune_stale_cache_files` deletes only matching sibling files.

### DIFF
--- a/src/dbt_bouncer/utils.py
+++ b/src/dbt_bouncer/utils.py
@@ -1,5 +1,6 @@
 """Re-usable functions for dbt-bouncer."""
 
+import contextlib
 import hashlib
 import importlib
 import importlib.util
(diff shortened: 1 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Fixes stale disk cache in editable installs by including internal check files in the registry cache fingerprint, and cleans up accumulated stale cache files.**

**Type:** `bugfix`

### Summary
In editable (development) installs the dbt-bouncer version is `"0.0.0"`, so adding or modifying an internal check file under `src/dbt_bouncer/checks/` did not change the cache fingerprint — the disk cache remained stale and forced a full-scan fallback on every run. This PR incorporates the internal `checks/` directory into the fingerprint hash. It also extracts a `_hash_py_tree()` helper to DRY up the hashing logic (previously only used for custom checks), and adds `_prune_stale_cache_files()` to clean up old `check_registry_*.json` files that accumulate when the fingerprint keeps changing.

### File Changes
| `src/dbt_bouncer/utils.py` | Modified | Add `_hash_py_tree()` helper, include internal `checks/` dir in fingerprint, add `_prune_stale_cache_files()` to clean up old cache files |
| `tests/unit/test_utils.py` | Modified | Add tests verifying fingerprint changes when internal checks are touched, and that stale cache files are pruned correctly |

### Review Findings
- ✅ No issues with **Python quality** — type hints present on all new function signatures, clean imports (`contextlib` added), proper docstrings with Args/Returns sections.
- ✅ No issues with **check structure** — no checks added/modified, only utility functions.
- ✅ No issues with **test coverage** — both new functions have dedicated tests covering the happy path. The fingerprint test verifies the before/after invariant and restores the original mtime in a `finally` block. The prune test covers keeping the active file, deleting stale files, and leaving unrelated files alone.
- ✅ No issues with **security** — no deserialization changes, no user input handling, `contextlib.suppress(OSError)` is appropriately scoped.
- ✅ No issues with **style** — consistent with existing code style, snake_case naming, sorted file iteration.
- ✅ No issues with **breaking changes** — all new/changed functions are private (`_`-prefixed), no public API or config schema changes.
- ✅ No issues with **documentation** — all new functions have complete docstrings. No architecture changes requiring AGENTS.md updates.
---

PR #884, opened 2026-05-12
perf: cache validated bouncer config to disk

pgoslatara:
After PR #883 lands, the remaining startup hot path on `dbt-bouncer run` is `validate_conf()`. It costs ~140 ms per run because Pydantic has to build a discriminated-union model spanning every check class referenced by the config, then validate the YAML against it. The result is deterministic for a given (config contents, packaged check files, custom check files, entry points, `DBT_PROJECT_DIR`) tuple, so this PR caches it to disk and skips the expensive part on the warm path.

The wrinkle is that the dynamically-built `DbtBouncerConf` subclass produced by `create_model()` cannot survive a process restart — pickle/JSON can't roundtrip the class itself. The workaround: cache a JSON payload listing the base field values plus `(_module, _qualname, model_dump())` for every check. On warm load, the relevant check modules are imported (so each class is resolvable), each check is rebuilt via `cls.model_construct(**data)`, and they're handed to a lightweight `DbtBouncerConfBase` subclass whose category fields are typed `list[Any]`. That subclass takes <1 ms to build vs ~70 ms for the discriminated-union variant, and `model_construct` skips validation since the data has already been validated once.

Two safety details:
- The cache is JSON (no pickle), so an attacker who can write the cache file cannot inject Python objects.
- `_resolve_check_class` only resolves names whose module is under `dbt_bouncer.checks.*` or already present in `sys.modules` (i.e. loaded by the normal `get_check_objects_for_names` pass), so a corrupted cache cannot pull in arbitrary third-party modules.

`DBT_BOUNCER_DISABLE_CONF_CACHE=1` opts out for tests or debugging.

Measured on `dbt-bouncer-example.yml` (533 checks): wall time goes from ~0.46 s cold to ~0.32 s warm — a ~30% reduction on the warm path that dominates real CI usage.

### DIFF
--- a/src/dbt_bouncer/configuration_file/validator.py
+++ b/src/dbt_bouncer/configuration_file/validator.py
@@ -20,17 +20,31 @@ if TYPE_CHECKING:
 
 _rebuilt_classes: set[str] = set()
 
-# Field names on DbtBouncerConfBase that we cache. The three ``_checks``
-# categories are added dynamically and handled separately.
-_BASE_FIELD_NAMES = (
-    "custom_checks_dir",
-    "dbt_artifacts_dir",
-    "exclude",
-    "include",
-    "package_name",
-    "severity",
-)
 _CHECK_CATEGORIES = ("catalog_checks", "manifest_checks", "run_results_checks")
+
+
+def _base_field_names() -> tuple[str, ...]:
+    """Return the names of the cacheable scalar fields on ``DbtBouncerConfBase``.
+
+    Derived from ``DbtBouncerConfBase.model_fields`` minus the three check
+    categories (which are dynamic and handled separately). Computing this on
+    demand means any new base field is picked up by the cache automatically —
+    no hand-maintained tuple to drift out of sync.
+
+    Returns:
+        tuple[str, ...]: Sorted field names to persist in the cache payload.
+
+    """
+    from dbt_bouncer.configuration_file.parser import DbtBouncerConfBase
+
+    return tuple(
+        sorted(
+            name
+            for name in DbtBouncerConfBase.model_fields
+            if name not in _CHECK_CATEGORIES
+        )
+    )
+
 
 DEFAULT_DBT_BOUNCER_CONFIG = """manifest_checks:
   - name: check_model_directories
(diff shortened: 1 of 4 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Caches the validated bouncer configuration to disk, avoiding expensive discriminated-union model rebuilds on repeated runs with the same config.**

**Type:** `perf`

### Summary
This PR introduces a disk-based cache for the validated `DbtBouncerConf` instance produced by `validate_conf()`. On the first ("cold") run, the fully-validated config is serialised to a JSON file under `~/.cache/dbt-bouncer/` keyed by a hash of the package version, check source files, config contents, and relevant environment variables. Subsequent ("warm") runs deserialise the cached payload via `model_construct()` (skipping re-validation) and resolve check classes through a strict allowlist built from the configured check names — avoiding the ~70ms cost of rebuilding Pydantic discriminated-union models. A companion refactor extracts `_internal_checks_digest()` with `lru_cache` so the packaged checks tree is hashed only once per interpreter session, benefiting both the existing registry cache and the new config cache.

<details>

| `src/dbt_bouncer/configuration_file/validator.py` | Modified | Adds config cache read/write logic (`_load_cached_conf`, `_write_cached_conf`), a lightweight Pydantic subclass factory (`_get_lite_conf_class`), cache toggle (`_conf_cache_enabled`), and integrates the cache into `validate_conf()` |
| `src/dbt_bouncer/utils.py` | Modified | Extracts `_internal_checks_digest()` with `lru_cache`, adds `compute_conf_cache_key()`, and extracts `get_cache_dir()` helper |
| `tests/unit/test_config_file_validator.py` | Modified | Adds 6 tests covering cold/warm paths, cache disable env var, corrupt cache fallback, custom checks dir, and cache key invalidation |
| `tests/unit/test_utils.py` | Modified | Updates existing fingerprint test to call `_internal_checks_digest.cache_clear()` between mutations |

</details>

### Review Findings

- ✅ No issues with **Python quality** — type hints are present on all new function signatures, imports are clean (lazy where needed to avoid circular deps), Pydantic `Field()` used correctly in `_get_lite_conf_class`.
- ✅ No issues with **check structure** — this PR does not add or modify any checks; the cache transparently wraps the existing validation path.
- ✅ No issues with **test coverage** — six new tests exercise the cold path (file written), warm path (matches cold), disable env var (no file written), corrupt/wrong-version cache (graceful fallback), custom checks directory, and cache key sensitivity to file mtime changes. Happy and unhappy paths are well covered.
- ✅ No issues with **security** — deserialisation uses `orjson.loads` (safe JSON, no arbitrary code execution). `_load_cached_conf` builds a strict `(module, qualname) -> class` allowlist from classes loaded for the user's configured check names; any cache entry referencing a class outside that set causes a full rebuild. No `pickle`, `yaml.unsafe_load`, or command injection vectors.
- ✅ No issues with **style** — snake_case naming throughout, `_CHECK_CATEGORIES` tuple is alphabetically ordered, new constants and helpers follow existing module conventions.
- ✅ No issues with **breaking changes** — the cache is transparent to callers of `validate_conf()` and opt-out via `DBT_BOUNCER_DISABLE_CONF_CACHE=1`. No public API signatures changed.
- ✅ No issues with **documentation** — all new public/internal functions have complete docstrings with Returns sections. `compute_conf_cache_key` includes a forward-looking note about env vars feeding `default_factory` callables.

---

PR #885, opened 2026-05-12
perf: avoid wasted work in runner at scale

pgoslatara:
While answering a question about scale behaviour I generated a synthetic config with 50,000 checks and ran it against the example dbt project. Wall time was ~10 s and a cProfile showed three concrete sources of waste in the runner.

**1. Deep-copying every check×resource pair upfront.** \`runner()\` did \`check.model_copy(deep=True)\` for every resource of the matching type *before* running \`_should_run_check\`, which filters on include/exclude/materialization/name — none of which need a copy to inspect. On the 50 k test that was 650 k Pydantic deep_copy calls, the vast majority thrown away because the include pattern didn't match. Reordering so the filter runs first cuts the 50 k case from ~10 s → ~3.8 s. Real configs benefit too: any check with a tight \`include\` regex now skips the copy for resources outside that include.

**2. Recomputing per-resource meta for every check.** \`_get_resource_meta(...)\` returns the \`dbt-bouncer.skip_checks\` meta entry for a resource. That depends only on the resource, but it was being recomputed for every (check, resource) pair (650 k times on the 50 k test). Caching the \`(resource, skip_checks)\` list per \`iterate_value\` and reusing it across checks drops another ~1.1 s (3.8 s → 2.7 s).

**3. Eager repr on a debug log.** \`logging.debug(f\"{bouncer_config=}\")\` evaluates the Pydantic repr of the whole config even when log level is INFO and the message is never emitted. With 50 k checks the recursive repr alone took ~0.5 s. Swapping to \`logging.debug(\"bouncer_config=%r\", bouncer_config)\` defers the repr to log-emission time (2.7 s → 2.2 s).

**Net effect:**

| Config | Before | After | Δ |
|---|---|---|---|
| 50 000 checks (synthetic) | ~10.0 s | ~2.2 s | **−78 %** |
| 533 checks (\`dbt-bouncer-example.yml\`) | 0.39 s | 0.37 s | small but consistent |

All 653 unit tests still pass. The runner output is unchanged because (a) the filtered-out copies were always going to be thrown away and (b) the cached meta is a pure function of the resource.

### DIFF
--- a/src/dbt_bouncer/cli/run/utils.py
+++ b/src/dbt_bouncer/cli/run/utils.py
@@ -196,7 +196,7 @@ def run_bouncer(
         config_file_contents=dict(config_file_contents),
         custom_checks_dir=Path(custom_checks_dir) if custom_checks_dir else None,
     )
-    logging.debug(f"{bouncer_config=}")
+    logging.debug("bouncer_config=%r", bouncer_config)
 
     for category in check_categories:
         if category in only_parsed:
@@ -233,7 +233,7 @@ def run_bouncer(
                 [c for c in getattr(bouncer_config, category) if c.name in check_names],
             )
 
-    logging.debug(f"{bouncer_config=}")
+    logging.debug("bouncer_config=%r", bouncer_config)
 
     dbt_artifacts_dir = Path(
         config_file_path.parent / (bouncer_config.dbt_artifacts_dir or "target")
--- a/src/dbt_bouncer/runner.py
+++ b/src/dbt_bouncer/runner.py
@@ -184,6 +184,25 @@ def runner(
     for check_category in ctx.check_categories:
         list_of_check_configs.extend(getattr(ctx.bouncer_config, check_category))
 
+    # Per-iterate_value cache of (resource, skip_checks_for_resource) tuples.
+    # The skip_checks meta depends only on the resource, so computing it once
+    # per resource (instead of once per (check, resource) pair) is a big win
+    # when the config has many checks targeting the same resource type.
+    resources_with_meta: dict[str, list[tuple[Any, list[str]]]] = {}
+
+    def _resources_for(iterate_value: str) -> list[tuple[Any, list[str]]]:
+        cached = resources_with_meta.get(iterate_value)
+        if cached is not None:
+            return cached
+        out: list[tuple[Any, list[str]]] = []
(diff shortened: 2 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Avoid redundant `model_copy(deep=True)` and `_get_resource_meta` calls in the runner hot loop, significantly reducing CPU and memory overhead at scale.**

**Type:** `refactor`

### Summary
The runner loop previously deep-copied every check *before* deciding whether it should run, and recomputed resource metadata for every (check, resource) pair even when many checks target the same resource type. This PR introduces two optimisations: (1) a per-`iterate_value` cache (`_resources_for`) that computes skip-checks metadata once per resource instead of once per (check, resource) pair, and (2) moving the `_should_run_check` filter *before* the `model_copy(deep=True)` call so skipped checks never allocate a copy. A minor logging hygiene fix in `utils.py` switches f-string debug logs to `%r`-style lazy formatting.

### File Changes
| `src/dbt_bouncer/runner.py` | Modified | Cache resource metadata per `iterate_value`; filter before deep-copying checks to avoid wasted allocations |
| `src/dbt_bouncer/cli/run/utils.py` | Modified | Replace f-string `logging.debug` calls with lazy `%r` formatting |

### Review Findings
- ✅ **Python Quality** — No issues. The new `_resources_for` closure is well-typed (`dict[str, list[tuple[Any, list[str]]]]`), imports are clean, and no dead code was introduced.
- ✅ **Correctness** — Passing the original `check` (instead of the deep-copied `check_i`) to `_should_run_check` is safe: `_should_run_check` only reads `check.include`, `check.exclude`, `check.materialization`, and `check.name` — all immutable config properties that are identical on the original and the copy.
- ✅ **Check Structure** — No changes to check decorators, `BaseCheck`, or `fail()` usage.
- ✅ **Test Coverage** — Pure performance refactor with no behavioural change; existing tests cover the runner logic.
- ✅ **Security** — No new deserialization, no secrets, no command injection vectors.
- ✅ **Style** — Lazy `%r` logging is the preferred pattern over f-strings in `logging.*` calls. Code is consistent with project conventions.
- ✅ **Breaking Changes** — None. No public API, config schema, or function signature changes.
- ✅ **Documentation** — Internal refactor; no public-facing doc updates needed.

---

PR #887, opened 2026-05-13
fix: re-coerce RootModel fields on warm conf-cache load

pgoslatara:
Running `dbt-bouncer --config-file dbt-bouncer-example.yml` on a populated conf cache surfaced 20 spurious WARN entries from the four `*_has_meta_keys` checks. Verbose logs showed each one crashing with `AttributeError: 'list' object has no attribute 'model_dump'`, which the executor's blanket `Exception` handler silently downgrades to WARN.

Root cause: the warm-cache loader in `validator._load_cached_conf` used `cls.model_construct(**item["data"])`, which skips Pydantic validation. The cache writer dumps each check via `model_dump(mode="json")`, so a `NestedDict` `RootModel` field is persisted as a raw list. On warm load, `keys` stayed a `list` and the check's `keys.model_dump()` blew up.

This switches the warm-path call to `cls.model_validate(...)` so RootModel fields are re-coerced back into typed instances. The cold path, the on-disk format, and the cache invalidation logic are unchanged.

Verified locally:

- `uv run pytest tests/unit` — 662 passed.
- `prek run --all-files` — clean.
- `uv run dbt-bouncer --config-file dbt-bouncer-example.yml` cold and warm — `SUCCESS=533 WARN=0 ERROR=0`.

Added a regression test (`test_validate_conf_warm_path_rehydrates_nested_dict_field`) that fails on `model_construct` and passes on `model_validate`.

### DIFF
--- a/src/dbt_bouncer/configuration_file/validator.py
+++ b/src/dbt_bouncer/configuration_file/validator.py
@@ -396,9 +396,11 @@ def _load_cached_conf(
     nothing outside that vouched-for set is reachable, so a corrupted cache
     file cannot pull in arbitrary modules.
 
-    Cached check instances are rebuilt via ``model_construct`` (no field
-    validation) because the cached payload was the output of a prior
-    successful Pydantic validation pass.
+    Cached check instances are rebuilt via ``model_validate`` so that
+    ``RootModel`` fields (e.g. ``NestedDict``) are re-coerced from their
+    JSON-mode dumps back into typed Pydantic instances. ``model_construct``
+    would skip that coercion and leave such fields as raw primitives, which
+    breaks any check that calls ``.model_dump()`` on them at runtime.
 
     Returns:
         The cached config, or ``None`` if the cache is missing, corrupt, or
@@ -452,7 +454,7 @@ def _load_cached_conf(
                     item["_qualname"],
                 )
                 return None
-            out.append(cls.model_construct(**item["data"]))
+            out.append(cls.model_validate(item["data"]))
         materialised[cat] = out
 
     return _get_lite_conf_class().model_construct(**materialised)
--- a/tests/unit/test_config_file_validator.py
+++ b/tests/unit/test_config_file_validator.py
@@ -653,6 +653,39 @@ def test_validate_conf_warm_path_matches_cold(isolated_cache_dir):  # noqa: ARG0
     assert warm.dbt_artifacts_dir == cold.dbt_artifacts_dir
 
 
+def test_validate_conf_warm_path_rehydrates_nested_dict_field(isolated_cache_dir):  # noqa: ARG001
+    """Warm-cache load must re-coerce RootModel fields (e.g. NestedDict).
+
+    Regression: a previous implementation used ``model_construct`` on the warm
+    path, which left ``keys`` as a raw ``list`` after JSON round-tripping and
+    broke any check that called ``.model_dump()`` on it.
+    """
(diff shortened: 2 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Fixes warm-cache deserialization of `RootModel` fields (e.g. `NestedDict`) by switching from `model_construct` to `model_validate`.**

**Type:** `bugfix`

### Summary
When loading check instances from the configuration cache (warm path), `model_construct` was used to rebuild Pydantic models. This skips field validation/coercion, so `RootModel`-based fields like `NestedDict` were left as raw Python primitives (e.g. plain `list`) instead of being re-coerced into their typed Pydantic wrappers. Any check that later called `.model_dump()` on such a field would break. The fix switches to `model_validate`, which runs full Pydantic coercion and correctly rehydrates nested types. A targeted regression test ensures the warm-path round-trip produces a proper `NestedDict` instance.

### File Changes
| `src/dbt_bouncer/configuration_file/validator.py` | Modified | Replaced `cls.model_construct(**item["data"])` with `cls.model_validate(item["data"])` and updated the docstring to explain why |
| `tests/unit/test_config_file_validator.py` | Modified | Added regression test `test_validate_conf_warm_path_rehydrates_nested_dict_field` verifying `NestedDict` survives a cache round-trip |

### Review Findings
- ✅ No issues with **Python quality** — the change is minimal and correct; `model_validate` accepts a dict (matching the `model_dump()` output stored in cache), docstring accurately explains the rationale.
- ✅ No issues with **check structure** — no check classes were added or modified.
- ✅ No issues with **test coverage** — the new test covers the exact regression scenario (cold write → warm read → assert `NestedDict` type and `.model_dump()` output). Happy path is tested; the unhappy path (raw `list` instead of `NestedDict`) is the implicit "before this fix" state.
- ✅ No issues with **security** — no new deserialization surface; `model_validate` is Pydantic's standard safe entry point, and the existing class allow-list (`class_map`) still gates which classes can be instantiated.
- ✅ No issues with **style** — consistent with surrounding code, naming follows project conventions.
- ✅ No issues with **breaking changes** — the cache format version is unchanged (appropriate, since the on-disk format didn't change — only the reconstruction method did). Existing caches will be read correctly.
- ✅ No issues with **documentation** — the in-code docstring was updated to reflect the new behavior.
- 🟡 **Suggestion:** `src/dbt_bouncer/configuration_file/validator.py:457` — `model_validate` runs full validation including potentially expensive validators on every cached check load, whereas `model_construct` was zero-cost. For configurations with many checks this could regress warm-path performance. Consider whether the perf benchmark (`make test-perf`) was run to confirm the impact is acceptable. If it's negligible this is fine as-is.
---

pgoslatara:
Re the 🟡 perf suggestion — measured locally on the warm conf-cache path with the `dbt-bouncer-example.yml` config (97 checks across all categories), 50 iterations per variant, 3 repeats:

| variant | mean / load |
| --- | --- |
| `model_validate` (this PR) | 7.7 – 7.9 ms |
| `model_construct` (previous) | 7.7 – 8.8 ms |

Delta lands in `±1 ms` (`±10%`) across runs — within measurement noise. Set against the full warm e2e of 708.05 ms reported by Bencher above (which itself didn't trip the regression threshold), the conf-cache load is ~1% of total runtime and the validation step is not a meaningful cost at this config size.

Conclusion: keeping `model_validate` as-is — the correctness fix has no measurable performance impact on the warm path.

PR #892, opened 2026-05-20
fix: allow for macros to be used in dbt projects definitions

mattcristal:
## What was done

Macros **can** be used in the dbt project yaml file.

Those are "valid" macro usage. 

We should account for those 😄

### DIFF
--- a/src/dbt_bouncer/checks/manifest/check_macros.py
+++ b/src/dbt_bouncer/checks/manifest/check_macros.py
@@ -1,3 +1,4 @@
+import os
 import re
 from pathlib import Path
 from typing import Annotated, Any, ClassVar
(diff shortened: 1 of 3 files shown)


pgoslatara:
@mattcristal 
* Do you have examples of macros used in `dbt_project.yml` that are flagged as unused by `check_macro_is_used`? I am aware of `on-run-start/end` macros and pre/post-hooks at the project level, but both of these are covered by the check.
* I'm going to think a bit more about this PR. Reason is that it reaches outside of the dbt artifacts and queries a file directly, that's a pattern that `dbt-bouncer` was intentionally designed to _not_ do, `dbt-bouncer` only requires access to artifacts (even if most artifact usage is alongside the repo files).

mattcristal:
@pgoslatara 

will close this for now

PR #893, opened 2026-05-20
feat: add check_seed_file_size

mattcristal:
## What was done

Yes to seeds, no to bloat.


pgoslatara:
@mattcristal 
* Same concern as mentioned [here](https://github.com/godatadriven/dbt-bouncer/pull/892#issuecomment-4507027066).
* Is there a way to do this via `catalog.json` and the `stats` key that is contained on each node? This differs per adapter but I know BigQuery has `num_rows` and `num_bytes`.

pgoslatara:
Closing this in favour of #898. Thanks for the idea for checking seed size!

PR #903, opened 2026-05-29
feat: add rust function for object in path

mattcristal:
## What was done

dbt-bouncer aims to:

> Be as **fast** as possible, running checks against dbt artifacts.

For this reason, I believe that there is some **room for improvement in some internal functions** / utilities for which we could use Rust 🦀 

Like in this case, there is indeed a noticeable **speed-up** benefit of 3x.

Python 🐍 & Rust 🦀  can work together very well, see Pydantic, Polars, and more.

⚠️  This PR is a bit **experimental**. It is not a light decision to introduce another language in the codebase, but it is good to try and **experiment**.

I am **open to feedback** about that and also about where the files are located (in this new PR).

What I like of this approach is that the change is self-contained. I am **starting small.**


mattcristal:
Relevant conversation:

https://github.com/godatadriven/dbt-bouncer/discussions/714

mattcristal:
@pgoslatara you can take a look when you have time

pgoslatara:
> Pádraic Slattery (@pgoslatara) you can take a look when you have time

@mattcristal Thanks for taking a look at Rust, as mentioned in the discussion you link to (#714) I think this can result in a valuable speed-up in terms of run times (and is the reason I approved the workflows on this PR, so we can see the actual speed-up). However the increased complexity of the codebase and the packaging requirements is something I still feel is not worth the trade-off (especially when I see well-run projects like SQLFluff [struggle](https://github.com/sqlfluff/sqlfluff/releases/tag/4.2.1) in this area). Maybe this will change in the future and we can add some Rust to the codebase, or maybe other Python packages will be created that solve the packaging complexity and we can benefit from these.

mattcristal:
> > Pádraic Slattery (Pádraic Slattery (@pgoslatara)) you can take a look when you have time
> 
> Matteo Renoldi (@mattcristal) Thanks for taking a look at Rust, as mentioned in the discussion you link to (#714) I think this can result in a valuable speed-up in terms of run times (and is the reason I approved the workflows on this PR, so we can see the actual speed-up). However the increased complexity of the codebase and the packaging requirements is something I still feel is not worth the trade-off (especially when I see well-run projects like SQLFluff [struggle](https://github.com/sqlfluff/sqlfluff/releases/tag/4.2.1) in this area). Maybe this will change in the future and we can add some Rust to the codebase, or maybe other Python packages will be created that solve the packaging complexity and we can benefit from these.

Thanks for taking a look Pádraic.

This was just experimental and something that I wanted to try out. I agree with your comment. Hopefully this is something that we will continue soon 🦀

PR #912, opened 2026-06-12
fix: surface Pydantic detail for config validation errors

pgoslatara:
When a config file failed validation for any reason other than an unknown check name — for example a wrong parameter type — `validate_conf` raised `RuntimeError("")` with no message, hiding all of Pydantic's detail from the user. Non-matching validation errors are now reported with their location and message, e.g.:

```
1. manifest_checks -> check_model_names -> model_name_pattern: Input should be a valid string
```

The closest-name suggestion is also guarded against an empty check registry, which previously raised `NameError` via an unbound variable instead of the intended error message.

### DIFF
--- a/src/dbt_bouncer/configuration_file/validator.py
+++ b/src/dbt_bouncer/configuration_file/validator.py
@@ -699,14 +699,20 @@ def validate_conf(
                 incorrect_name = error["msg"][
                     error["msg"].find("tag") + 5 : error["msg"].find("found using") - 2
                 ]
-                min_dist = 100
-                for name in accepted_names:
-                    dist = jellyfish.levenshtein_distance(name, incorrect_name)
-                    if dist < min_dist:
-                        min_dist = dist
-                        min_name = name
+                min_name = min(
+                    accepted_names,
+                    key=lambda name,
+                    target=incorrect_name: jellyfish.levenshtein_distance(name, target),
+                    default=None,
+                )
+                suggestion = f" Did you mean '{min_name}'?" if min_name else ""
                 error_message.append(
-                    f"{len(error_message) + 1}. Check '{incorrect_name}' does not match any of the expected checks. Did you mean '{min_name}'?"
+                    f"{len(error_message) + 1}. Check '{incorrect_name}' does not match any of the expected checks.{suggestion}"
+                )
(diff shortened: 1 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Surfaces Pydantic validation details for config errors that don't match the "unknown check name" pattern, instead of silently dropping them.**

**Type:** `bugfix`

### Summary
Previously, when a `ValidationError` contained errors other than an unrecognised check name (e.g. a wrong parameter type), the `else` branch was missing entirely — those errors were silently swallowed, producing an empty or incomplete error message. This PR adds an `else` clause that formats arbitrary Pydantic errors with their field location and message, and refactors the Levenshtein suggestion lookup to use `min()` with a `default` for safety. A new test exercises the previously-uncovered unhappy path.

### File Changes
| `src/dbt_bouncer/configuration_file/validator.py` | Modified | Added `else` branch to surface non-name-match Pydantic errors; refactored Levenshtein nearest-name search to use `min()` with `default=None` |
| `tests/unit/test_config_file_validator.py` | Modified | Added `test_validate_conf_invalid_parameter_type` covering the new `else` branch with a wrong-type parameter scenario |

### Review Findings
- ✅ No issues with **Python quality** — the `min()` refactor is cleaner and the `default=None` guard is appropriate; lambda default-arg capture of `target` avoids the classic closure-in-loop bug.
- ✅ No issues with **check structure** — changes are in the config validator, not in a check class; no decorator or `BaseCheck` concerns apply.
- ✅ No issues with **test coverage** — the new test verifies the error message structure (numbered prefix, field location components, Pydantic message) for the previously-untested `else` path. The existing `test_validate_conf_incorrect_names` covers the `if` branch. Both happy and unhappy paths are covered.
- ✅ No issues with **security** — no deserialization, no secrets, no injection vectors.
- ✅ No issues with **style** — snake_case naming, formatting looks ruff-compliant, consistent with surrounding code.
- ✅ No issues with **breaking changes** — error messages are diagnostic output, not part of a programmatic API; the change is additive (new `else` branch) and the existing `if` branch output is unchanged.
- ✅ No issues with **documentation** — this is an internal bugfix; no public API or architecture change.

---

PR #914, opened 2026-06-12
fix: derive PR metadata from trusted workflow_run payload

pgoslatara:
The AI review and benchmark tracking workflows run with elevated privileges (`pull-requests: write`, `PAT_GITHUB`/`BENCHER_API_TOKEN` secrets) but read the PR number, head ref and SHAs from an `event.json` artifact uploaded by the untrusted PR run. A fork PR could forge that artifact to redirect review comments and Bencher reports — or the checked-out "PR head" — to an arbitrary PR or SHA.

Both workflows now resolve the PR through the GitHub API from `workflow_run.head_sha`, which GitHub populates server-side and cannot be forged by the PR author, and additionally gate on `workflow_run.event == 'pull_request'`. If the PR head has moved since the CI run, the job fails cleanly rather than acting on stale metadata (the newer CI run re-triggers it). The `event.json` upload in the CI pipeline had no other consumers and is removed.

While in the file, the Bencher API token is moved from a CLI argument to the `BENCHER_API_TOKEN` environment variable, which Bencher reads natively — CLI arguments are visible to other processes on the runner.

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -163,12 +163,6 @@ jobs:
               name: benchmark_results.json
               path: ./benchmark_results.json
 
-          - name: Upload GitHub Pull Request Event
-            uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
-            with:
-              name: event.json
-              path: ${{ github.event_path }}
-
     dev-container:
       needs: [prek]
       runs-on: ubuntu-24.04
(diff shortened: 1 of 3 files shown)


pgoslatara:
Note: the failed "Pull Request AI Review" / "Pull Request Benchmarks Track" runs triggered by this PR's CI are a one-time transition artifact — this PR's CI pipeline (correctly) no longer uploads `event.json`, but until this PR merges the `workflow_run` consumers still execute the artifact-based definition from `main`. They self-heal on merge.

PR #915, opened 2026-06-12
perf: skip dbt-bouncer's own entry points during plugin discovery

pgoslatara:
dbt-bouncer registers its three check packages under the `dbt_bouncer.checks` entry point group, and `_load_entry_point_checks` loaded them on every discovery call. Via `pkgutil.walk_packages` this imported every internal check module even when `get_check_objects_for_names` only needed one — nullifying the targeted-import fast path that the disk-cached module map exists to provide.

Internal checks are already discovered through the filesystem scan and the cached name-to-module map, so the entry-point loader now skips any entry point resolving into `dbt_bouncer.checks`. Third-party plugin entry points are unaffected. Measured with a warm module-map cache: a config naming a single check now imports 3 check modules instead of 43.

The `pyproject.toml` registrations are deliberately left in place — they are harmless for external discoverability and removing them would not protect environments with an older installed distribution.

### DIFF
--- a/src/dbt_bouncer/utils.py
+++ b/src/dbt_bouncer/utils.py
@@ -283,6 +283,15 @@ def _load_entry_point_checks(check_objects: list[type["BaseCheck"]]) -> None:
     """
     eps = entry_points(group=_ENTRY_POINT_GROUP)
     for ep in eps:
+        # dbt-bouncer registers its own check packages in this group; loading
+        # them here would import every internal check module, defeating the
+        # targeted-import fast path. Internal checks are discovered via the
+        # filesystem scan / module map instead.
+        if ep.module == "dbt_bouncer.checks" or ep.module.startswith(
+            "dbt_bouncer.checks."
+        ):
+            logging.debug(f"Skipping internal entry point `{ep.name}`.")
+            continue
         try:
             target = ep.load()
 
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -415,7 +415,9 @@ class CheckEntryPointFake(BaseCheck):
         """No-op execute for testing."""
 
 
-def _make_fake_ep(name: str, target: object) -> MagicMock:
+def _make_fake_ep(
+    name: str, target: object, module: str = "fake_plugin.checks"
+) -> MagicMock:
     """Create a fake entry point that loads to the given target.
 
     Returns:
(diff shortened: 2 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Skip dbt-bouncer's own entry points during plugin discovery to avoid redundant imports and preserve the targeted-import fast path.**

**Type:** `perf`

### Summary
dbt-bouncer registers its internal check packages (e.g. `dbt_bouncer.checks.catalog`, `dbt_bouncer.checks.manifest`) as entry points. Before this change, `_load_entry_point_checks` would eagerly load all of them — importing every internal check module — even though internal checks are already discovered via the filesystem scan / module map. This PR adds an early `continue` to skip any entry point whose `module` starts with `dbt_bouncer.checks`, avoiding the redundant import overhead.

The test file is updated in lockstep: the `_make_fake_ep` helper gains a `module` parameter so tests can simulate internal entry points, and a new test asserts that those entry points are never loaded.

### File Changes
| `src/dbt_bouncer/utils.py` | Modified | Skip entry points under `dbt_bouncer.checks` in `_load_entry_point_checks` |
| `tests/unit/test_utils.py` | Modified | Add `module` param to `_make_fake_ep`; add test for internal-EP skipping |

### Review Findings

<details>
<summary>All clear — expand for details</summary>

- ✅ **Python Quality**: Type hints are consistent with existing code; no dead code introduced; clean conditional logic.
- ✅ **Check Structure**: Not applicable — no new checks added.
- ✅ **Test Coverage**: New test (`test_load_entry_point_checks_skips_internal_entry_points`) covers the happy path (all three internal EP modules skipped, `load()` never called, `check_objects` remains empty). Existing tests continue to exercise the non-skipped path via the default `module="fake_plugin.checks"`.
- ✅ **Security**: No deserialization, no secrets, no user-controlled input in the new code path.
- ✅ **Style**: f-string in `logging.debug` is consistent with surrounding code and `G004` is in the project's ruff ignore list. Formatting matches ruff-format output.
- ✅ **Breaking Changes**: None — external plugins whose entry-point modules are outside `dbt_bouncer.checks.*` are unaffected. Internal behaviour only.
- ✅ **Documentation**: Comment in `utils.py` clearly explains *why* internal EPs are skipped. No public API change, so no doc update needed.
- ✅ **Cross-file version consistency**: Not applicable — no version bumps in this PR.

</details>

---

PR #916, opened 2026-06-12
fix: gate release pipeline on tests and reorder publish steps

pgoslatara:
The release pipeline pushed git tags, then Docker images, then PyPI, then the GitHub release — all in one job with no validation. A failure at the PyPI step left `vX`/`vX.X`/`vX.X.X` tags and Docker images already published with no package behind them, and nothing verified the build before tagging.

Two changes:

1. **Test gate** — the unit test suite runs after the version bump and before any publish step.
2. **PyPI first, idempotent rest** — PyPI is published before tags/images/release: it is the canonical artifact (the Docker image's entrypoint installs `dbt-bouncer==X` from PyPI at runtime, so an image without its PyPI package is broken anyway) and the only step that cannot be re-pointed afterwards. Every subsequent step is now idempotent — `skip-existing: true` on the PyPI publish, tags were already force-pushed, and `gh release create` is guarded by a `gh release view` check — so a partial failure is recoverable by simply re-running the workflow with the same inputs instead of stranding a half-published release.

I considered splitting into `needs`-gated jobs, but that requires shuttling the un-pushed version-bumped tree between jobs as artifacts; reordering plus idempotent steps achieves the same recoverability with less moving machinery. `dist_pypi/` is gitignored, so building the wheel before the `git add -A` release commit does not pollute it.

### DIFF
--- a/.github/workflows/release_pipeline.yml
+++ b/.github/workflows/release_pipeline.yml
@@ -115,11 +115,19 @@
                   git add -A
                   git commit -m "Bumping version to v${STEPS_VERSION_OUTPUTS_VERSION}"
 
-                  # Tag and push X.X.X
-                  git tag -f \
-                    -a v${STEPS_VERSION_OUTPUTS_VERSION} \
-                    -m "v${STEPS_VERSION_OUTPUTS_VERSION}"
-                  git push -f origin "v${STEPS_VERSION_OUTPUTS_VERSION}"
+                  # Tag and push X.X.X — skipped if it already exists on the
+                  # remote (a re-run after a partial failure) so a published
+                  # version tag is never moved to a new commit. The alias
+                  # tags below are intentionally mutable and always
+                  # (re-)pushed.
+                  if git ls-remote --exit-code --tags origin "refs/tags/v${STEPS_VERSION_OUTPUTS_VERSION}" > /dev/null; then
+                    echo "Tag v${STEPS_VERSION_OUTPUTS_VERSION} already exists on origin, not re-tagging."
+                  else
+                    git tag -f \
+                      -a v${STEPS_VERSION_OUTPUTS_VERSION} \
+                      -m "v${STEPS_VERSION_OUTPUTS_VERSION}"
+                    git push -f origin "v${STEPS_VERSION_OUTPUTS_VERSION}"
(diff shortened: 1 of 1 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Reorders and hardens the release pipeline so PyPI publishing happens earlier and idempotent guards prevent duplicate tags/releases on re-runs.**

**Type:** `chore`

### Summary
This PR restructures the release pipeline to move the PyPI build and publish steps earlier in the workflow (before tagging/pushing), and adds idempotent guards so the pipeline can be safely re-run without failing on already-existing tags or releases. The exact-version tag (`vX.X.X`) no longer force-pushes; instead it checks for existence first and skips if present. A unit test step is also added before publishing. Additionally, `skip-existing: true` is added to the PyPI publish action for extra safety.

### File Changes
| `.github/workflows/release_pipeline.yml` | Modified | Reorders PyPI publish before tagging, adds idempotent guards for tags and releases, adds unit test step, removes force-push for exact version tags |

### Review Findings
- 🟡 **Suggestion:** `.github/workflows/release_pipeline.yml:113` — The minor-version tag (`X.X`) still uses `git tag -f` and `git push -f`, meaning re-runs will force-update it. This is likely intentional (floating tag), but worth confirming it's the desired asymmetry vs. the exact-version tag which now skips if it exists.
- ✅ No issues with **Security** — no secrets exposed, no unsafe deserialization, actions pinned to commit SHAs.
- ✅ No issues with **Breaking Changes** — no config schema or public API changes.
- ✅ No issues with **Cross-file version consistency** — the `pypa/gh-action-pypi-publish` action hash is unchanged and only referenced in this one file.
- ✅ No issues with **Style** — YAML formatting is consistent.
- ✅ No issues with **Python Quality / Check Structure / Test Coverage / Documentation** — this PR only touches CI workflow YAML; no Python code or docs changed.
---

pgoslatara:
Re the AI review finding: added a guard so the `vX.X.X` tag is never re-tagged if it already exists on the remote (a re-run after a partial failure no longer moves a published version tag). The `vX` / `vX.X` alias tags remain intentionally mutable — they must point at the latest release of their line, which on a re-run is this one.

PR #919, opened 2026-06-18
feat: support dbt-core 2.0 (Fusion)

pgoslatara:
Makes dbt-bouncer support dbt-core 2.0 (the Rust Fusion engine) by making the fixture project build under both dbt 1.12 and 2.0, adding a committed `dbt_20` fixture, and parameterising the tests over it. No `src/dbt_bouncer/**` changes: dbt 2.0 emits the same schema-v12 artifacts that the parser already reads.

## Fixture project

`dbt_version_at_least(target)` is a new gate macro, used to no-op `apply_grants` under dbt 2.0 where the bundled implementation is incompatible with the duckdb adapter. The Python model is disabled under 2.0 (duckdb Python models are unsupported there) via the inline `dbt_version` built-in rather than the macro — project macros are not in scope when dbt evaluates `enabled:` at parse time.

Semantic models move from the standalone `_semantic_models.yml` to the nested `semantic_model:` spec on the model itself. Both live build versions parse it (each reports 1 metric, 1 semantic model), so no version gate is needed.

## Build matrix

`build-artifacts-20` is ported from the old `makefile` to `mise.toml`. It must not use `dbt parse`: a parse-only run on dbt 2.0 emits **no** macro-to-macro `depends_on` edges at all (0/1076 across every package, project macros included), which silently breaks macro-graph checks such as `check_macro_is_used`.

Both version tasks now run `dbt build` first. `catalog.json` is produced by introspecting the DuckDB database and `./dbt.duckdb` is gitignored, so the previous recipes depended on ambient local state: run from a fresh clone they emitted an **empty** catalog and every catalog check passed vacuously. Fixing this took the example config from 610 to 712 passing checks on `dbt_20`, and 638 to 744 on `dbt_112` — roughly 100 catalog checks per fixture that previously never fired.

dbt 2.0 needs three steps, because no single command writes all three artifacts: `build` populates the database and writes `manifest.json` plus a real `run_results.json`, then `compile --write-catalog` writes `catalog.json` into a scratch target path (it also rewrites `run_results.json` with zero results, so only the catalog is kept). `build-artifacts-local` moves to `compile --write-catalog`, since the venv now resolves 2.0.

dbt 1.10 and 1.11 are archived alongside 1.7/1.8: their committed fixtures stay for backward-compat parsing coverage but are no longer rebuilt, because they cannot parse the 1.12+ semantic spec and schema YAML cannot be version-gated with `{% if %}`. The active build matrix is 1.12 + 2.0.

The dev dependency becomes `dbt-core>=2.0.0a5,<3` with `prerelease = "allow"`, so the canonical venv runs 2.0 directly. 1.x coverage is preserved through the uvx-isolated `build-artifacts-112` build.

## Verification

| | result |
|---|---|
| Full example config vs the dbt 2.0.0a5 fixture | `SUCCESS=712 WARN=0 ERROR=0` |
| Full example config vs the dbt 1.12 fixture | `SUCCESS=744 WARN=0 ERROR=0` |
| `dbt_112` fixture vs `main` | catalog 16 nodes, 58 run results — unchanged |
| `dbt_20` fixture | catalog 15 nodes, 62 run results |
| Unit + integration tests | 1578 passed |
| `prek run --all-files` | clean |

## Notes for review

Rebasing the original branch was impractical — it was 139 commits behind `main` and conflicted on its first commit — so its net change was re-applied onto current `main` as three commits. Three consequences worth flagging:

- **`check_macro_is_used` no longer needs a workaround.** The previous branch added `exclude: ^macros/system/` to `dbt-bouncer-example.yml`; #1044 fixes the underlying detection in `src`, so the exclude is dropped and the check passes on the new fixture unaided.
- **The `e2e-dbt-20` CI job is dropped as redundant.** With the dev pin on 2.0, the existing test-matrix job already runs `mise run build-and-run-dbt-bouncer` against a 2.0 venv on every OS/Python combination; a dedicated job would run the identical command. Happy to re-add it purely as a named, legible check.
- **Two new `check_macro_has_meta_keys` failures** appeared because `main` began requiring `meta.owner` on macros after this branch was written; both new macros now carry it.

The committed `dbt_20` fixture tracks the same three JSON artifacts as `dbt_112` (`manifest.json`, `catalog.json`, `run_results.json`); dbt 2.0's extra per-node output directories are gitignored.


pgoslatara:
## 🔍 AI PR Review
> **Adds dbt 2.0 (Fusion) support: new test fixtures, updated build pipeline, semantic model migration, and documentation.**

**Type:** `docs`

### Summary
This PR adds dbt 2.0 (Fusion) compatibility to dbt-bouncer. dbt 2.0 drops `dbt docs generate` and changes how artifacts are produced, so the build pipeline now uses `dbt compile --write-catalog` and `dbt build` as separate steps. The test fixture project migrates from standalone `_semantic_models.yml` to the inline `semantic_model:` spec (dbt 1.12+ syntax), adds workaround macros for Fusion incompatibilities (`apply_grants`, `dbt_version_at_least`), disables Python models on dbt 2.0, and commits new `dbt_20` fixtures. Test coverage is widened: parser tests now parametrise across all four fixture versions (1.10, 1.11, 1.12, 2.0), and the integration failure-path test covers both the oldest frozen and newest fixture. Documentation and the SECURITY.md version matrix are updated accordingly.

The dbt 1.10 and 1.11 build tasks are removed from `mise.toml` (their fixtures are frozen but remain committed), and the dev dependency is bumped to `dbt-core>=2.0.0a5,<3` with `prerelease = "allow"`.

<details>
<summary>📄 File Changes (25 files)</summary>

| `.agents/skills/build-artifacts/SKILL.md` | Modified | Updated to reference dbt 1.12 + 2.0 fixtures; improved verification instructions to check catalog node count |
| `.gitignore` | Modified | Added ignore patterns for dbt 2.0 extra target directories and dbt 1.12 `run/` directory |
| `AGENTS.md` | Modified | Updated `build-artifacts` description from "(dbt 1.10, 1.11, 1.12)" to "(dbt 1.12, 2.0)" |
| `SECURITY.md` | Modified | Added dbt 2.0 (Fusion) row to version support matrix |
| `dbt_project/macros/system/_system__macros.yml` | Modified | Added YAML documentation for new `apply_grants` and `dbt_version_at_least` macros |
| `dbt_project/macros/system/apply_grants.sql` | New | Override macro that skips grant application on dbt 2.0 (incompatible with duckdb adapter) |
| `dbt_project/macros/system/dbt_version_at_least.sql` | New | Helper macro for version comparison using `major*100 + minor` encoding |
| `dbt_project/models/marts/finance/_finance__models.yml` | Modified | Migrated semantic model definition inline (entities, dimensions, metrics on `orders` model) |
| `dbt_project/models/marts/finance/_semantic_models.yml` | Deleted | Standalone semantic model file replaced by inline spec in `_finance__models.yml` |
| `dbt_project/models/utilities/_utilities__models.yml` | Modified | Disabled `py_order_stats` Python model on dbt 2.0 (unsupported on duckdb) |
| `docs/faq.md` | Modified | Added FAQ section about dbt 2.0 / Fusion compatibility |
| `docs/getting_started.md` | Modified | Updated artifact generation instructions to recommend `dbt compile --write-catalog`; added caveat about empty `run_results.json` |
| `mise.toml` | Modified | Removed dbt 1.10/1.11 build tasks; added `build-artifacts-20` task; updated 1.12 to run `dbt build` before `docs generate`; updated `build-artifacts-local` to use `compile --write-catalog` |
| `pyproject.toml` | Modified | Bumped dev `dbt-core` to `>=2.0.0a5,<3`; added `prerelease = "allow"` to `[tool.uv]` |
| `schema.json` | Modified | Updated `CheckMacroIsUsed` description with dbt 2.0 `dbt parse` caveat |
| `src/dbt_bouncer/checks/manifest/check_macros.py` | Modified | Added `!!! warning` admonition about `dbt parse` vs `dbt compile` on dbt 2.0 for `check_macro_is_used` |
| `tests/fixtures/dbt_112/target/catalog.json` | Modified | Regenerated fixture (now built with `dbt build` first so catalog has real data) |
| `tests/fixtures/dbt_112/target/manifest.json` | Modified | Regenerated fixture reflecting new macros and model configs |
| `tests/fixtures/dbt_112/target/run_results.json` | Modified | Regenerated fixture |
| `tests/fixtures/dbt_20/target/catalog.json` | New | New dbt 2.0 catalog fixture |
| `tests/fixtures/dbt_20/target/manifest.json` | New | New dbt 2.0 manifest fixture |
| `tests/fixtures/dbt_20/target/run_results.json` | New | New dbt 2.0 run results fixture |
| `tests/integration/test_programmatic_invocation.py` | Modified | Parametrised `test_programmatic_failure_path` over `dbt_110` and `dbt_20` fixtures |
| `tests/unit/test_parsers_manifest.py` | Modified | Added shared `dbt_artifacts_dir` fixture parametrised across all 4 fixture versions; both parser tests now use it |
| `uv.lock` | Modified | Lockfile updated for dbt-core 2.0 alpha dependency |

</details>

### Review Findings

- 🟡 **Suggestion:** `docs/getting_started.md:6` — The updated instructions recommend `dbt compile --write-catalog` as the primary path, but the same paragraph says "On dbt 1.x, use `dbt docs generate`". Since dbt-bouncer v4 supports dbt 1.7 through 2.0 (per SECURITY.md), it may be clearer to lead with the dbt 1.x command (still the majority of users) and note the 2.0 alternative, or restructure as a version-conditional list. Minor clarity nit.

- 🟡 **Suggestion:** `tests/unit/test_parsers_manifest.py` — The `dbt_artifacts_dir` fixture is module-scoped by default (no `scope` set on `@pytest.fixture`). Since the fixture only constructs a `Path` object this is fine functionally, but adding `scope="session"` would make the intent explicit and match the project's convention of session-scoped fixtures noted in AGENTS.md.

- 🟡 **Suggestion:** `dbt_project/models/utilities/_utilities__models.yml:10` — The `enabled:` Jinja expression `{{ (dbt_version.split('.')[0] | int) < 2 }}` is a good workaround with a clear comment explaining why `dbt_version_at_least()` cannot be used here. However, the expression evaluates to a string (`"True"` / `"False"`) in Jinja — dbt treats both as truthy strings. Verify this actually disables the model on dbt 2.0 in the fixture (the `py_order_stats` model should be absent from the dbt_20 manifest). If dbt coerces the string correctly this is fine, but it's worth a sanity check.

- ✅ No issues with **Python quality** — no new Python code beyond test parametrisation, which is clean and well-structured.
- ✅ No issues with **check structure** — the only check change (`check_macro_is_used`) is a docstring/warning update, no logic changes.
- ✅ No issues with **test coverage** — parser tests now cover all 4 fixture versions; integration test covers oldest + newest. The `check_macro_is_used` warning is documentation-only so no new test cases needed.
- ✅ No issues with **security** — no unsafe deserialization, no secrets, no command injection. Jinja macros use safe string operations.
- ✅ No issues with **style** — macro YAML entries are alphabetically ordered, `.gitignore` entries are logically grouped.
- ✅ No issues with **breaking changes** — no config schema changes, no removed checks, no changed function signatures. The `dbt-core` bump is dev-only.
- ✅ No issues with **documentation** — AGENTS.md updated, SECURITY.md updated, FAQ added, getting_started updated.
- ✅ No issues with **cross-file version consistency** — `dbt docs generate` references in `mise.toml` are all in comments explaining the migration; no stale action references found in CI workflows.

---

Issue #925, opened 2026-06-30
[Bug] check_column_has_specified_test false positives on case-folded catalog column names (Snowflake)


sti0:
### Current Behavior

`check_column_has_specified_test` (v3.5.0) produces false positives on warehouses
that fold catalog identifiers to a fixed case (e.g. Snowflake → UPPERCASE).

The check compares **catalog** column names against the `column_name` of tests
parsed from **manifest.json**. These two sources have different casing:

- `catalog.json` column names are UPPERCASE (`CREATED_AT`, `LOAD_DATETIME`, …)
- test `column_name` in `manifest.json` mirrors the dbt YAML, conventionally
  lowercase (`created_at`, `load_datetime`, …)

The presence check is a plain, case-sensitive set-membership comparison, so an
UPPERCASE catalog column never matches its lowercase test entry. Every targeted
column is flagged as missing the test even though the test exists.

Root cause — `dbt_bouncer/checks/catalog/columns/tests.py`:

```python
columns_to_check = [
    v.name                                   # catalog column name -> UPPERCASE on Snowflake
    for _, v in catalog_node.columns.items()
    if compiled_column_name_pattern.match(str(v.name)) is not None
]
tested_columns = set()
for t in ctx.tests:
    ...
    tested_columns.add(getattr(t, "column_name", ""))   # manifest/YAML name -> lowercase

non_complying_columns = [c for c in columns_to_check if c not in tested_columns]
```

`"CREATED_AT" not in {"created_at"}` is always `True`. The `column_name_pattern`
regex only filters which catalog columns enter `columns_to_check`; it does not
affect the final comparison, and `compile_pattern` is called with `flags=0`
(no `re.IGNORECASE`). There is no config option to make the comparison
case-insensitive.

### Expected Behavior

The check should pass when a matching test exists regardless of identifier casing.
For a column `CREATED_AT` in the catalog and a `not_null` test declared on
`created_at` in the YAML, the check should report success.

Suggested fix — case-fold both sides of the presence comparison:

```python
tested_columns = {c.casefold() for c in tested_columns}
non_complying_columns = [
    c for c in columns_to_check if c.casefold() not in tested_columns
]
```

(The same casing mismatch may affect other catalog checks that cross-reference
manifest column names — worth a sweep.)

### Steps To Reproduce

1. Materialise a model on Snowflake whose catalog column is upper-cased, e.g.
   `my_model` with a `CREATED_AT` column.

[...]

PR #926, opened 2026-06-30
fix: case-insensitive column matching in check_column_has_specified_test (#925)

pgoslatara:
`check_column_has_specified_test` produced false positives on warehouses that fold catalog identifiers to a fixed case (e.g. Snowflake → UPPERCASE), as reported in #925. Catalog column names (`CREATED_AT`) were compared case-sensitively against the lowercase `column_name` from manifest.json (`created_at`), so every targeted column was wrongly flagged as untested.

The fix mirrors the existing `check_columns_are_all_documented` pattern: a `case_sensitive` parameter (default `true`) that auto-relaxes to `false` for the `snowflake` adapter and lower-cases both sides of the comparison. The `column_name_pattern` regex is also compiled with `re.IGNORECASE` in that mode, otherwise a lowercase pattern (`^is_.*`) never matches an upper-cased column and the check passes vacuously. Model-level tests (no `column_name`) are skipped rather than added as empty entries.

I preferred this over the issue's unconditional `casefold()` because case-insensitive matching is wrong on case-sensitive adapters with quoted identifiers (BigQuery, quoted Postgres).

Both catalog checks exposing `case_sensitive` are also aligned with AGENTS.md: the parameter is documented under `Parameters` (not `Receives`), and its type is tightened from `bool | None = True` to `bool = True` since `None` was never a meaningful state.

Tests cover Snowflake auto-fold, explicit `case_sensitive=False`, the case-sensitive default still failing, and a lowercase pattern matching an upper-cased Snowflake column. `schema.json` regenerated.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -104,7 +104,7 @@
     },
     "CheckColumnHasSpecifiedTest": {
       "additionalProperties": false,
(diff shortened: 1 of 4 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Adds case-insensitive column matching to `check_column_has_specified_test` for Snowflake compatibility, aligning it with the existing `check_columns_are_all_documented` pattern.**

**Type:** `bugfix`

### Summary
On Snowflake, catalog columns are upper-cased (e.g. `IS_ACTIVE`) while manifest test entries use the lowercase YAML name (`is_active`). The existing `check_column_has_specified_test` performed case-sensitive matching, so columns on Snowflake silently fell out of scope — the regex never matched the upper-cased name, causing the check to pass vacuously instead of flagging a missing test.

This PR adds a `case_sensitive` parameter (defaulting to `true`, auto-set to `false` for Snowflake) and applies `re.IGNORECASE` to the pattern compilation when case-insensitive. It also refactors `check_columns_are_all_documented` to tighten the `case_sensitive` type from `bool | None` to `bool`, and moves the parameter from the `Receives` docstring section to `Parameters` to match AGENTS.md conventions. The schema.json is regenerated to reflect both changes.

<details>

| `src/dbt_bouncer/checks/catalog/columns/tests.py` | Modified | Add `case_sensitive` param, case-insensitive regex compilation, and case-folded column comparison |
| `src/dbt_bouncer/checks/catalog/columns/description.py` | Modified | Tighten `case_sensitive` type from `bool \| None` to `bool`; move it from Receives to Parameters in docstring |
| `tests/unit/checks/catalog/columns/test_tests.py` | Modified | Add 4 new test cases covering Snowflake auto-detection, explicit case-insensitive, case-sensitive mismatch, and pattern-vs-uppercase-column |
| `schema.json` | Modified | Regenerated to add `case_sensitive` to `CheckColumnHasSpecifiedTest` and tighten type in `CheckColumnsAreAllDocumented` |

</details>

### Review Findings

- ✅ No issues with **Python quality** — type hints are correct, `bool | None` properly tightened to `bool`, `import re` added where needed, no dead code.
- ✅ No issues with **check structure** — proper use of `@check` decorator, `fail()` for failures, `ctx` for manifest access.
- ✅ No issues with **test coverage** — 4 new tests cover: Snowflake auto-detection (happy), explicit `case_sensitive=False` (happy), case-sensitive mismatch (unhappy), and lowercase pattern vs uppercase catalog column on Snowflake (unhappy). Both `check_passes` and `check_fails` paths are exercised.
- ✅ No issues with **security** — no deserialization changes, no secrets, no command injection vectors.
- ✅ No issues with **style** — parameters listed alphabetically in docstrings, snake_case naming, clean formatting.
- ✅ No issues with **breaking changes** — `case_sensitive` defaults to `true`, preserving existing behavior for non-Snowflake adapters. The `bool | None` to `bool` tightening in `check_columns_are_all_documented` is non-breaking since `None` was never a meaningful user-provided value (the default was `True`).
- ✅ No issues with **documentation** — docstrings correctly updated with Parameters/Receives/Other Parameters sections per AGENTS.md conventions; `case_sensitive` properly moved from Receives to Parameters since it's user-configurable.

---

PR #928, opened 2026-06-30
fix: parse test-macro args from signature, not body (#927)

### LINKED ISSUE #927: [Bug] Test-macro argument parser fails on {% set %} statements ('Assign' object has no   attribute 'nodes')
### Current Behavior

`check_macro_arguments_description_populated` (v3.5.1) crashes with an internal
`AttributeError` for any **test macro** (`{% test ... %}`) whose body contains a
`{% set ... %}` statement before the test call.

dbt-bouncer downgrades the crash to a `WARN`, so the check silently stops
validating that macro's arguments instead of either passing or failing.

Root cause — `dbt_bouncer/checks/manifest/check_macros.py`
(`check_macro_arguments_description_populated`):

```python
ast = environment.parse(macro.macro_sql)

if hasattr(ast.body[0], "args"):
    # "true" macro
    macro_arguments = [a.name for a in getattr(ast.body[0], "args", [])]
else:
    if "materialization" in [...]:
        macro_arguments = []
    else:
        # Macro is a test
        test_macro = next(
            x
            for x in ast.body
            if not isinstance(x.nodes[0], nodes.Call)   # <-- assumes x has `.nodes`
        )
        macro_arguments = [
            x.name
            for x in test_macro.nodes
            if isinstance(x, nodes.Name)
        ]
```

The test branch iterates `ast.body` and assumes every top-level node exposes
`.nodes`. A `{% set %}` statement parses to a `jinja2.nodes.Assign`, which has
`.node` / `.target` but **no** `.nodes` attribute, so `x.nodes[0]` raises
`AttributeError: 'Assign' object has no attribute 'nodes'`.

### Expected Behavior

The argument parser should tolerate non-`Output` top-level nodes (e.g. `Assign`
from `{% set %}`) and still extract the test macro's arguments, so the check
passes when the arguments are documented and fails when they are not — without
crashing.

Suggested fix — guard the `.nodes` access (skip nodes that don't expose it):

```python
test_macro = next(
    x
    for x in ast.body
    if getattr(x, "nodes", None) and not isinstance(x.nodes[0], nodes.Call)
)
```

### Steps To Reproduce

1. Define a test macro that uses `{% set %}` before the test expression —
   `macros/expect_valid_length.sql`:

   ```jinja
   {% test expect_valid_length(model, column_name, list_values) %}
   {%- set expression = column_name ~ " in (" ~ list_values | join(',') ~ ")" -%}
   {{ dbt_expectations.expression_is_true(model, expression=expression) }}
   {% endtest %}
   ```

2. Document the test macro and its arguments — `macros/expect_valid_length.yml`:

   ```yaml
   version: 2
   macros:
     - name: test_expect_valid_length
       description: Asserts the column value is in the allowed list.
       arguments:
         - name: model
           description: The model under test.
         - name: column_name
           description: The column to check.
         - name: list_values
           description: Allowed values.
   ```

3. Configure the check — `dbt-bouncer.yml`:

   ```yaml
   manifest_checks:
     - name: check_macro_arguments_description_populated
   ```

4. Generate artifacts and run:

   ```bash
   dbt parse
   dbt-bouncer --config-file dbt-bouncer.yml
   ```

### Relevant log output

```shell
`dbt-bouncer` encountered an error (AttributeError: 'Assign' object has no
attribute 'nodes'. Did you mean: 'node'?), run with `-v` to see more details or
report an issue at https://github.com/godatadriven/dbt-bouncer/issues.

Done. SUCCESS=... WARN=1 ERROR=0


The arguments are fully documented, but the `{% set %}` statement in the test
macro body crashes the AST walk, so the check never validates them.
```

### Environment

```markdown
- OS: Linux
- Python: 3.12.12
- dbt-bouncer: 3.5.1
```

### Add your company's website here to be added to our list of [users](https://godatadriven.github.io/dbt-bouncer/stable/users/#users).

_No response_

pgoslatara:
`check_macro_arguments_description_populated` crashed with `AttributeError: 'Assign' object has no attribute 'nodes'` for any test macro whose body contained a `{% set %}` statement before the test expression, as reported in #927. The crash was downgraded to a `WARN`, so the check silently stopped validating that macro's arguments instead of passing or failing.

The previous logic walked the macro body looking for the first top-level node whose `nodes[0]` was not a `Call`. That heuristic is fragile: a `{% set %}` statement parses to a `jinja2.nodes.Assign`, which has no `.nodes` attribute, and a test body that is purely a single macro call (e.g. `{{ dbt_expectations.expression_is_true(...) }}`) has no template node to harvest argument names from at all. The one-line guard suggested in the issue avoids the `AttributeError` but then matches no node and raises `StopIteration`, so it does not actually fix the reporter's example.

This change extracts the argument names directly from the test signature (`{% test name(arg_1, arg_2, ...) %}`), unwrapping the autoescape `escape()` wrapper to reach the test's own call node. The signature is the source of truth for which arguments exist, so the parser now works for every body shape — template SQL, a single macro call, and bodies containing `{% set %}` statements.

Added unit cases covering a fully-documented test macro with a `{% set %}` body (passes) and the same macro missing an argument description (fails). The existing materialization branch is unchanged.

Closes #927

### DIFF
--- a/src/dbt_bouncer/checks/manifest/check_macros.py
+++ b/src/dbt_bouncer/checks/manifest/check_macros.py
@@ -95,16 +95,19 @@ def check_macro_arguments_description_populated(
             # Materializations don't have arguments
             macro_arguments = []
         else:
-            # Macro is a test
-            test_macro = next(
-                x
-                for x in ast.body
-                if not isinstance(x.nodes[0], nodes.Call)  # type: ignore[attr-defined]
-            )
+            # Macro is a test. Extract the argument names from the test
+            # signature (`{% test name(arg_1, arg_2, ...) %}`) rather than
+            # walking the macro body. Walking the body is fragile: it breaks
+            # for tests whose body is a single macro call or contains a
+            # `{% set %}` statement, which parses to a `jinja2.nodes.Assign`
+            # node with no `.nodes` attribute (see issue #927).
+            signature_call = ast.body[0].nodes[0]  # type: ignore[attr-defined]
+            # With autoescape enabled the signature is wrapped in an
+            # `escape(...)` call; unwrap it to reach the test's own call node.
+            if signature_call.args and isinstance(signature_call.args[0], nodes.Call):
+                signature_call = signature_call.args[0]
             macro_arguments = [
-                x.name
-                for x in test_macro.nodes  # type: ignore[attr-defined]
-                if isinstance(x, nodes.Name)
+                a.name for a in signature_call.args if isinstance(a, nodes.Name)
             ]
 
     # macro_arguments: List of args parsed from macro SQL
--- a/tests/unit/checks/manifest/test_macros.py
+++ b/tests/unit/checks/manifest/test_macros.py
@@ -75,6 +75,39 @@ from dbt_bouncer.testing import check_fails, check_passes
             check_fails,
             id="missing_argument_in_config",
         ),
+        pytest.param(
+            {
+                "arguments": [
+                    {"name": "model", "description": "The model under test."},
+                    {"name": "column_name", "description": "The column to check."},
+                    {"name": "list_values", "description": "Allowed values."},
+                ],
+                # Test macro whose body is a single macro call: the `{% set %}`
+                # statement parses to a `jinja2.nodes.Assign` that has no `.nodes`
+                # attribute (see issue #927).
+                "macro_sql": '{% test expect_valid_length(model, column_name, list_values) %}\n{%- set expression = column_name ~ " in (" ~ list_values | join(\',\') ~ ")" -%}\n{{ dbt_expectations.expression_is_true(model, expression=expression) }}\n{% endtest %}',
+                "name": "test_expect_valid_length",
+                "original_file_path": "macros/expect_valid_length.sql",
+                "unique_id": "macro.package_name.test_expect_valid_length",
+            },
+            check_passes,
+            id="test_macro_with_set_statement",
+        ),
+        pytest.param(
+            {
+                "arguments": [
+                    {"name": "model", "description": "The model under test."},
+                    {"name": "column_name", "description": ""},
+                    {"name": "list_values", "description": "Allowed values."},
+                ],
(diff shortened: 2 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Fixes test-macro argument parsing by reading args from the Jinja signature instead of walking the macro body, preventing crashes on `{% set %}` statements.**

**Type:** `bugfix`

### Summary
The previous approach to extracting test-macro arguments iterated over the macro body looking for `nodes.Name` nodes, but this broke when the body contained a `{% set %}` statement (which parses to a `jinja2.nodes.Assign` node lacking a `.nodes` attribute). The fix extracts argument names directly from the test's signature call node (`{% test name(arg1, arg2) %}`), which is both simpler and more robust. Two new parametrised test cases (happy + unhappy) exercise the exact macro pattern that triggered the bug.

### File Changes
| `src/dbt_bouncer/checks/manifest/check_macros.py` | Modified | Replaced body-walking logic with signature-based argument extraction for test macros, with autoescape unwrapping |
| `tests/unit/checks/manifest/test_macros.py` | Modified | Added two parametrised test cases for test macros containing `{% set %}` statements (pass and fail paths) |

### Review Findings
- ✅ No issues with **Python quality** — type ignore comments are appropriate for dynamic Jinja2 AST access, no dead code introduced.
- ✅ No issues with **check structure** — existing `@check` decorator and `fail()` usage unchanged; only internal parsing logic modified.
- ✅ No issues with **test coverage** — both happy path (`test_macro_with_set_statement`) and unhappy path (`test_macro_with_set_statement_missing_description`) added, covering the exact regression case from issue #927.
- ✅ No issues with **security** — no deserialization changes, no user-controlled input used unsafely.
- ✅ No issues with **style** — code follows existing patterns, comment explains the *why* clearly with issue reference.
- ✅ No issues with **breaking changes** — internal implementation detail only; no config schema, function signature, or public API changes.
- ✅ No issues with **documentation** — inline comments explain the autoescape unwrapping and link to the issue; no public API change requiring docstring updates.

---

PR #948, opened 2026-07-07
fix(ci): skip AI review cleanly when a PR head is stale

pgoslatara:
The `Pull Request AI Review` workflow (`pull_request_ai_review.yml`) is triggered by `workflow_run` and resolves the PR from the triggering CI run's head SHA. When the PR branch has moved on since that CI run (new commits pushed), `listPullRequestsAssociatedWithCommit` finds no open PR at that SHA, so `Export PR Event Data` returned early via `core.setFailed` without exporting `PR_HEAD_SHA` / `PR_BASE` / `PR_NUMBER`. Because `Post or update review comment` is declared `if: always()`, it then ran with those variables empty — `git diff "origin/..."` fell through to its usage text and `gh api .../issues//comments` returned `Not Found (HTTP 404)`, turning the run red (e.g. run 28797856218).

This makes a stale/superseded head a benign skip rather than a failure: `Export PR Event Data` now emits `core.warning` and sets a `skip=true` step output, and the `Checkout PR Head`, `Analyse PR`, and `Post or update review comment` steps are gated on `steps.export.outputs.skip != 'true'`. A stale run ends green (all steps skipped); the fresh CI run for the new head still triggers a full review. Genuine errors in the export step continue to fail as before.

PR #949, opened 2026-07-07
fix(ci): skip benchmark tracking cleanly when a PR head is stale

pgoslatara:
Companion to #948. The `Pull Request Benchmarks Track` workflow (`pull_request_benchmarks_track.yml`) is a second `workflow_run` consumer of `CI pipeline` and contains the **same copy-pasted `Export PR Event Data` step** that #948 fixed in the AI-review workflow. When a PR branch is updated (rebase/force-push) after its CI run started, the in-flight CI run completes with a now-stale head, no open PR matches it, and `core.setFailed` turns this workflow red (e.g. run 28848587093, stale head `efe0d3c…`).

This applies the identical neutral-skip guard: `Export PR Event Data` now emits `core.warning` + sets `skip=true`, and the `bencher` install and `Track Benchmarks with Bencher` steps are gated on `steps.export.outputs.skip != 'true'`. A stale run ends green (steps skipped); the fresh CI run for the current head still tracks benchmarks normally.

Note: the root cause is that this `Export PR Event Data` block is duplicated across both `workflow_run` workflows — a good follow-up would be to extract it into a shared composite action so this class of fix lives in one place.

### DIFF
--- a/.github/workflows/pull_request_benchmarks_track.yml
+++ b/.github/workflows/pull_request_benchmarks_track.yml
@@ -23,6 +23,7 @@ jobs:
           run_id: ${{ github.event.workflow_run.id }}
 
       - name: Export PR Event Data
+        id: export
         uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
         with:
           script: |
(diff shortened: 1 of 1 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Prevents confusing red CI runs when a PR's head SHA becomes stale before the benchmark-tracking workflow fires.**

**Type:** `bugfix`

### Summary
When new commits are pushed to a PR after a CI run starts but before the `workflow_run`-triggered benchmark tracking job executes, the job could not find an open PR matching the old head SHA and hard-failed with `core.setFailed()`. This surfaced a misleading red check on an already-superseded run. The fix downgrades the failure to a warning, sets a `skip` output, and uses `if:` conditions to skip the Bencher steps cleanly — the fresh CI run for the new head will produce its own benchmark tracking.

### File Changes
| `.github/workflows/pull_request_benchmarks_track.yml` | Modified | Replace `core.setFailed()` with `core.warning()` + skip output; gate Bencher steps behind `steps.export.outputs.skip != 'true'` |

### Review Findings
- ✅ No issues with **Python quality** (no Python files changed)
- ✅ No issues with **check structure** (no check files changed)
- ✅ No issues with **test coverage** (CI workflow change; not unit-testable)
- ✅ No issues with **security** (no secrets handling changes; existing `BENCHER_API_TOKEN` usage is unchanged)
- ✅ No issues with **style** (YAML is clean, comments are clear and explain the *why*)
- ✅ No issues with **breaking changes** (behavior change is strictly more lenient — stale runs skip instead of fail)
- ✅ No issues with **documentation** (inline comments adequately explain the rationale)
- ✅ No issues with **cross-file version consistency** (`bencherdev/bencher@50fb1e138651a46d2fb704fab1adab38c181552e # v0.6.6` is unchanged and matches `merge_pipeline.yml:32`)
---

PR #950, opened 2026-07-07
refactor(ci): extract shared PR resolver into a composite action

pgoslatara:
Both `pull_request_ai_review.yml` and `pull_request_benchmarks_track.yml` are `workflow_run` consumers of `CI pipeline` and carried an **identical copy of the `Export PR Event Data` step** — the block that resolves the PR from the run's head SHA and skips cleanly when that head is stale. That duplication is exactly why the stale-head bug had to be patched twice (#948 for the AI review, #949 for the benchmark).

This extracts it into a local composite action, `.github/actions/resolve-pr-from-workflow-run`, which exports the PR metadata (`PR_HEAD`, `PR_HEAD_SHA`, `PR_BASE`, `PR_BASE_SHA`, `PR_NUMBER`, `PR_CHANGED_FILES`) as environment variables and exposes a `skip` output. Both workflows now sparse-check-out the action into a subdir and call it, keeping their existing `steps.export.outputs.skip != 'true'` gating. This is a pure refactor with no behaviour change — the stale-head handling now lives in one place.

Because `workflow_run` workflows can't be exercised on the PR itself, the end-to-end behaviour is validated on the first AI-review/benchmark run after merge.

### DIFF
--- a/.github/actions/resolve-pr-from-workflow-run/action.yml
+++ b/.github/actions/resolve-pr-from-workflow-run/action.yml
@@ -1,0 +1,55 @@
+name: Resolve PR from workflow_run
+description: >-
+  Resolve the pull request for a workflow_run event's head SHA and export its
+  metadata (PR_HEAD, PR_HEAD_SHA, PR_BASE, PR_BASE_SHA, PR_NUMBER,
+  PR_CHANGED_FILES) as environment variables for subsequent steps in the job.
+  If no open PR matches the head SHA (the branch was updated since the
+  triggering CI run), the `skip` output is set to 'true' instead of failing,
+  so callers can no-op cleanly rather than erroring on stale data.
+
+outputs:
+  skip:
+    description: >-
+      'true' when no open PR matches the workflow_run head SHA (a stale run);
+      otherwise 'false'. Gate downstream steps on this.
+    value: ${{ steps.resolve.outputs.skip }}
+
+runs:
+  using: composite
+  steps:
+    - id: resolve
+      uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+      with:
+        script: |
+          // Derive PR metadata from the trusted workflow_run payload — an
+          // artifact uploaded by the untrusted PR run could be forged to point
+          // a privileged workflow at an arbitrary PR/SHA.
+          const headSha = context.payload.workflow_run.head_sha;
+          const { data: prs } = await github.rest.repos.listPullRequestsAssociatedWithCommit({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            commit_sha: headSha,
+          });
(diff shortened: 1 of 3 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Extracts duplicated PR-resolution logic from two workflow files into a shared composite action, reducing maintenance burden and ensuring consistent behavior.**

**Type:** `refactor`

### Summary
Both `pull_request_ai_review.yml` and `pull_request_benchmarks_track.yml` contained identical inline `actions/github-script` blocks to resolve a PR from a `workflow_run` event's head SHA, handle stale runs, and export PR metadata as env vars. This PR extracts that logic into a new reusable composite action at `.github/actions/resolve-pr-from-workflow-run/action.yml` and replaces the inline blocks with `uses:` references. The composite action unifies a minor behavioral difference: the benchmark workflow previously did not export `PR_CHANGED_FILES`, but the shared action now exports it for all callers (harmless — it's only consumed by the AI-review prompt). A preceding docs commit clarifies this in the action's description.

### File Changes
| `.github/actions/resolve-pr-from-workflow-run/action.yml` | New | Reusable composite action that resolves a PR from `workflow_run` head SHA, exports PR metadata env vars, and outputs `skip=true` for stale runs |
| `.github/workflows/pull_request_ai_review.yml` | Modified | Replaces inline github-script PR resolution with a checkout of the composite action + `uses:` reference |
| `.github/workflows/pull_request_benchmarks_track.yml` | Modified | Same replacement as the AI-review workflow; adds sparse checkout step for the composite action |

### Review Findings
- ✅ No issues with **security** — the composite action correctly derives PR metadata from the trusted `workflow_run` payload (not from artifacts), and both callers use `persist-credentials: false` on the sparse checkout.
- ✅ No issues with **cross-file version consistency** — `actions/checkout` pin (`9c091bb2...`) and `actions/github-script` pin (`3a2844b7...`) match all other occurrences in the repo. No stale references to the old inline script remain.
- ✅ No issues with **breaking changes** — both workflows preserve the same `steps.export.outputs.skip` gate and the same env var names (`PR_HEAD`, `PR_HEAD_SHA`, `PR_BASE`, `PR_BASE_SHA`, `PR_NUMBER`, `PR_CHANGED_FILES`). The benchmark workflow now additionally exports `PR_CHANGED_FILES`, which is unused but harmless, and documented in a code comment.
- ✅ No issues with **style** — comments are clear, YAML is well-structured, the `_pr_resolver` checkout path convention is consistent between both callers.
- ✅ No issues with **documentation** — the composite action has a thorough `description`, the `skip` output is documented, and inline code comments explain design choices (e.g., why `PR_CHANGED_FILES` is exported for all callers).
- ✅ No issues with Python quality, check structure, or tests (not applicable — this PR touches only GitHub Actions YAML/JS).
---

PR #966, opened 2026-07-14
fix: skip deprecation-shim module during check discovery

pgoslatara:
Running the test suite surfaced five identical `DeprecationWarning`s in the pytest summary, each attributed to an unrelated check test:

```
DeprecationWarning: Importing from 'dbt_bouncer.checks.common' is deprecated.
Use 'from dbt_bouncer.check_framework.exceptions import ...' instead.
```

The root cause is that check discovery in `utils.py` globs `checks/**/*.py` and imports every module it finds. That sweep included `checks/common.py` — a backward-compatibility shim that fires a module-level `warnings.warn(..., DeprecationWarning)` on import. So the internal discovery pass tripped the warning that was only ever meant for external callers still importing `dbt_bouncer.checks.common`. The shim contributes no `Check*` classes, so importing it was pure dead weight. `get_check_objects` even had an explicit `# Always include top-level .py files (e.g. common.py)` comment deliberately pulling it in.

The count of exactly five is an artefact of `--numprocesses 5`: each pytest-xdist worker imported the shim once (subsequent imports are cached in `sys.modules`), and each attributed its single warning to whichever test first triggered discovery in that worker.

Both discovery loops (`_build_check_module_map` and `get_check_objects`) now skip any filename in a new `_NON_CHECK_MODULES` set (currently just `common.py`). A parametrised regression test asserts neither discovery entry point imports the shim.

The shim itself is unchanged and still emits its warning for genuine external imports (covered by the existing `test_deprecation_shims.py`). The full unit + integration suites pass with zero warnings.

### DIFF
--- a/src/dbt_bouncer/utils.py
+++ b/src/dbt_bouncer/utils.py
@@ -348,6 +348,12 @@ _CATEGORY_TO_SUBDIR: dict[str, str] = {
 
 _SUBDIR_TO_CATEGORY: dict[str, str] = {v: k for k, v in _CATEGORY_TO_SUBDIR.items()}
 
+# Modules that live under ``checks/`` but define no check classes. These are
+# skipped during discovery so their import side effects don't fire — notably
+# ``common.py`` is a backward-compat shim that emits a DeprecationWarning on
+# import (see ``checks/common.py``).
+_NON_CHECK_MODULES: set[str] = {"common.py"}
+
 
 def _build_check_module_map() -> dict[str, dict[str, str]]:
     """Build a mapping of check_name -> {module, category} by scanning check modules.
@@ -364,6 +370,9 @@ def _build_check_module_map() -> dict[str, dict[str, str]]:
     mapping: dict[str, dict[str, str]] = {}
 
     for check_file in (f for f in checks_dir.glob("**/*.py") if f.is_file()):
+        if check_file.name in _NON_CHECK_MODULES:
+            continue
+
         index = check_file.parts.index("checks")
         module_name = ".".join(
             ["dbt_bouncer", "checks", *check_file.parts[index + 1 :]]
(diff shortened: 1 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Skip the backward-compat deprecation shim (`checks/common.py`) during check discovery to prevent spurious `DeprecationWarning`s.**

**Type:** `bugfix`

### Summary
The `checks/common.py` module is a backward-compatibility shim that re-exports symbols from `check_framework.exceptions` and emits a `DeprecationWarning` at import time. Both check-discovery functions (`_build_check_module_map` and `get_check_objects`) glob `**/*.py` under `checks/`, which caused them to import this shim and trigger the warning even though it contains no check classes. This PR introduces a `_NON_CHECK_MODULES` set to skip known non-check modules during discovery, and adds a parametrised test verifying neither discovery path triggers the deprecation warning.

### File Changes
| `src/dbt_bouncer/utils.py` | Modified | Add `_NON_CHECK_MODULES` set and skip listed modules in both `_build_check_module_map` and `get_check_objects` |
| `tests/unit/test_utils.py` | Modified | Add parametrised test asserting that check discovery does not import the deprecation shim |

### Review Findings
- ✅ No issues with **Python quality** — clean constant definition, proper type hint (`set[str]`), clear docstring comment.
- ✅ No issues with **test coverage** — test covers both discovery paths (`_build_check_module_map` and `get_check_objects`) and correctly uses `recwarn` to detect stray warnings. Cache is cleared before and after to ensure isolation.
- ✅ No issues with **security** — no deserialization, no user input handling.
- ✅ No issues with **style** — consistent with existing code patterns, comment style matches codebase conventions.
- ✅ No issues with **breaking changes** — `common.py` is still importable for external consumers; it's only skipped during internal discovery.
- ✅ No issues with **documentation** — the inline comment on `_NON_CHECK_MODULES` clearly explains the rationale.
---

PR #975, opened 2026-07-17
fix: escape rich markup in console failure report

### LINKED ISSUE #974: Console failure messages strip [...] (e.g. regex character classes) because rich markup is enabled
## Summary

Failure messages printed to the console have `rich` markup **enabled**, so any
text that looks like a `rich` tag — most importantly regex character classes
like `[a-z0-9]` — is interpreted as markup and silently stripped. The result is
that `check_model_names` (and any check whose failure message contains `[...]`)
prints a regex that is **not** the one actually being enforced, which is
misleading when debugging a failing check.

Matching itself is unaffected — this is purely a display bug in the console
report. The JUnit/CSV/JSON outputs are correct because they don't go through
`rich`.

## Reproduction

Config:

```yaml
manifest_checks:
  - name: check_model_names
    include: models
    model_name_pattern: ^[a-z0-9]+(_[a-z0-9]+)*__[a-z0-9]+(_[a-z0-9]+)*$
```

For a model that fails the check, the console "Failed checks" table prints:

```
`my_model` does not match the supplied regex `^+(_+)*__+(_+)*$`.
```

**Expected:**

```
`my_model` does not match the supplied regex `^[a-z0-9]+(_[a-z0-9]+)*__[a-z0-9]+(_[a-z0-9]+)*$`.
```

Every `[a-z0-9]` has been removed because `rich` parsed `[a-z0-9]` as a style
tag.

### Minimal root-cause repro (no dbt-bouncer needed)

```python
from rich.console import Console
from rich.table import Table
from rich.markup import escape

c = Console(width=80)
t = Table()
t.add_column("Failure message")
msg = "`x` does not match the supplied regex `^[a-z0-9]+(_[a-z0-9]+)*$`."
t.add_row(msg)          # -> `x` does not match the supplied regex `^+(_+)*$`.
t.add_row(escape(msg))  # -> `x` does not match the supplied regex `^[a-z0-9]+(_[a-z0-9]+)*$`.
c.print(t)
```

## Root cause

`dbt_bouncer/reporting/reporter.py` adds the raw failure message to the table
cell, where `rich` interprets it as markup:

```python
table.add_row(
    str(check.get("check_run_id", "")),
    f"[bold {sev_color}]{sev.upper()}[/bold {sev_color}]",
    str(check.get("failure_message", "")),   # <- markup interpreted here
)
```

## Suggested fix

Escape the user-derived content before handing it to `rich` (the severity cell
is intentional markup and should stay as-is):

```python
from rich.markup import escape
...
table.add_row(
    escape(str(check.get("check_run_id", ""))),
    f"[bold {sev_color}]{sev.upper()}[/bold {sev_color}]",
    escape(str(check.get("failure_message", ""))),
)
```

(Escaping `check_run_id` too guards against model/resource names that contain
`[...]`.)

## Environment

- dbt-bouncer 3.7.0
- rich 14.3.4
- Python 3.13.9

pgoslatara:
Failure messages printed to the console had `rich` markup enabled, so any text that looks like a `rich` tag — most importantly regex character classes such as `[a-z0-9]` — was interpreted as markup and silently stripped. As a result `check_model_names` (and any check whose failure message contains `[...]`) printed a regex that was **not** the one actually being enforced, which was misleading when debugging a failing check. Matching behaviour was unaffected, and the JUnit/CSV/JSON outputs were already correct because they don't route through `rich`.

The fix escapes the user-derived cells (`check_run_id`, `file_path`, and `failure_message`) with `rich.markup.escape` before adding them to the table in `reporting/reporter.py`. The severity cell is intentional markup and is left as-is. A regression test in `tests/unit/reporting/test_reporter.py` renders a failure message containing `[a-z0-9]` and asserts the character class survives in the console output.

Fixes #974

### DIFF
--- a/src/dbt_bouncer/reporting/reporter.py
+++ b/src/dbt_bouncer/reporting/reporter.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING, Any
 
 from rich import box
 from rich.console import Console
+from rich.markup import escape
 from rich.panel import Panel
 from rich.table import Table
 
@@ -183,11 +184,15 @@ class Reporter:
                 sev = str(check.get("severity", "")).lower()
                 sev_color = "red" if CheckSeverity.ERROR in sev else "yellow"
 
+                # Escape user-derived content so text that looks like rich
+                # markup (e.g. regex character classes such as `[a-z0-9]`) is
+                # not interpreted as a style tag and stripped. The severity
+                # cell is intentional markup and is left as-is. See issue #974.
                 table.add_row(
-                    str(check.get("check_run_id", "")),
-                    str(check.get("file_path") or ""),
+                    escape(str(check.get("check_run_id", ""))),
+                    escape(str(check.get("file_path") or "")),
                     f"[bold {sev_color}]{sev.upper()}[/bold {sev_color}]",
-                    str(check.get("failure_message", "")),
+                    escape(str(check.get("failure_message", ""))),
                 )
 
             console.print(table)
--- a/tests/unit/reporting/test_reporter.py
+++ b/tests/unit/reporting/test_reporter.py
@@ -111,6 +111,34 @@ def test_report_results_saves_coverage_file(tmp_path):
     assert output_file.exists()
 
 
+def test_report_results_escapes_rich_markup_in_console(capsys, monkeypatch):
+    """Failure messages containing `[...]` are not stripped by rich markup.
+
+    Regression test for a display bug where regex character classes such as
+    `[a-z0-9]` in a failure message were interpreted as rich style tags and
+    silently removed from the console report (see issue #974).
+    """
+    # Force a wide console so the failure message renders on a single line and
+    # the pattern isn't split across rich's column wrapping.
+    monkeypatch.setenv("COLUMNS", "250")
+
+    pattern = "^[a-z0-9]+(_[a-z0-9]+)*$"
+    results = [
+        {
+            "check_run_id": "check_model_names:0:model.my_model",
+            "failure_message": f"`my_model` does not match the supplied regex `{pattern}`.",
+            "file_path": "models/staging/my_model.sql",
+            "outcome": CheckOutcome.FAILED,
+            "severity": CheckSeverity.ERROR,
+        },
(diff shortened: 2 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Escape Rich markup in console failure reports so regex patterns like `[a-z0-9]` are no longer silently stripped from output.**

**Type:** `bugfix`

### Summary
When dbt-bouncer printed check failures to the console, user-derived strings (check name, file path, failure message) were passed directly to Rich's `Table.add_row()`. Rich interprets square-bracket sequences as markup style tags, so content like regex character classes (`[a-z0-9]`) was silently consumed and removed from the rendered output. This PR wraps those three user-derived cells with `rich.markup.escape()` while leaving the severity cell — which uses intentional Rich markup for colour — unchanged. A regression test confirms the fix.

### File Changes
| `src/dbt_bouncer/reporting/reporter.py` | Modified | Import `rich.markup.escape` and wrap user-derived table cell values to prevent Rich from interpreting them as style tags |
| `tests/unit/reporting/test_reporter.py` | Modified | Add regression test verifying that `[a-z0-9]` in a failure message survives console rendering |

### Review Findings
- ✅ No issues with **Python quality** — new import is clean, `escape()` usage is correct and targeted, type consistency preserved.
- ✅ No issues with **check structure** — this PR does not add or modify any dbt-bouncer checks.
- ✅ No issues with **test coverage** — the new test covers the exact regression scenario (bracket-containing failure message), sets a wide console to avoid wrapping artifacts, and asserts the literal `[a-z0-9]` substring survives in captured stdout.
- ✅ No issues with **security** — `escape()` is a hardening measure; no unsafe deserialization or injection vectors introduced.
- ✅ No issues with **style** — import is in alphabetical order within the `rich` block, snake_case naming, inline comment references the upstream issue.
- ✅ No issues with **breaking changes** — only the rendered console output changes (previously-stripped text now appears); no API, config schema, or function signature changes.
- ✅ No issues with **documentation** — the inline comment explaining the escape rationale and issue reference is sufficient for a localised bugfix.
---

PR #977, opened 2026-07-20
feat: utilize jinja library lexer to neutralize jinja

mattcristal:
## What was done

src/dbt_bouncer/sql_utils.py — neutralize_jinja now walks jinja2's own lexer instead of four regexes

### Why it's more idiomatic

The repo **already ships jinja2** and already parses Jinja with the real library in `check_macros`.py. This consolidates the two Jinja strategies into one and fixes real correctness gaps the regex couldn't handle — most notably {% set x = "%}" %}, where the non-greedy \{%.*?%\} truncated at the wrong %} and corrupted the SQL.

### Performance

- Perf (the accepted trade-off, measured): per unique model, neutralize went from ~1 µs → ~21 µs — but it's now ~15% of the neutralize+parse pipeline (sqlglot dominates at ~85%), it's @lru_cached so repeat calls are ~0.03 µs, and both structural checks now share one cache entry per model instead of neutralizing twice. Absolute cost ≈ 21 ms per 1,000 unique models — negligible.

### DIFF
--- a/src/dbt_bouncer/sql_utils.py
+++ b/src/dbt_bouncer/sql_utils.py
@@ -42,6 +42,10 @@ def _get_jinja_lexer_environment() -> "Environment":
     """
     from jinja2 import Environment
 
+    # ``autoescape`` only affects rendering, and this environment is used solely
+    # for its lexer (``.lex``), so it has no functional effect here. It is kept
+    # as a defensive default to satisfy bandit's B701 (jinja2_autoescape_false),
+    # matching ``check_macros`` which builds its environment the same way.
     return Environment(autoescape=True)
 
 
--- a/src/dbt_bouncer/sql_utils.py
+++ b/src/dbt_bouncer/sql_utils.py
@@ -10,50 +10,108 @@ to fall back to a best-effort approach.
 
 import re
 from functools import lru_cache
-from typing import cast
+from typing import TYPE_CHECKING, cast
 
 import sqlglot
 from sqlglot import exp
 from sqlglot.errors import SqlglotError
 
-# Constructs that render to nothing in real dbt are dropped so the surrounding
-# SQL still parses: comments (``{# ... #}``), statements (``{% ... %}`` - set,
-# if/for/macro control flow) and ``{{ config(...) }}``. Remaining ``{{ ... }}``
-# expressions occupy a value/table position, so each becomes a single-part
-# placeholder identifier - notably ``{{ ref('x') }}`` never survives as a
-# dotted (hard-coded-looking) reference.
-# Public so the regex-fallback path in ``checks.manifest.models.code`` can reuse
-# the same definition of a Jinja comment.
+if TYPE_CHECKING:
+    from jinja2 import Environment
+
+# ``JINJA_COMMENT_PATTERN`` stays public so the regex-fallback path in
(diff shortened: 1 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Replaces fragile regex-based Jinja neutralization with jinja2's own lexer, correctly handling edge cases like delimiters inside strings and nested braces.**

**Type:** `docs` (commit message says `docs: clarify lexer-only autoescape and test unlexable-input contract`, though the substance is `refactor` + `test`)

### Summary
The previous `neutralize_jinja` implementation used non-greedy regexes to strip Jinja constructs from model SQL before parsing with sqlglot. This broke on real-world edge cases: a `%}` inside a Jinja string literal would prematurely close the block, nested braces in dict expressions would be split incorrectly, and whitespace-control variants (`{%- ... -%}`) weren't explicitly handled. This PR replaces the four regex patterns with a token-stream walk over jinja2's own lexer, which understands Jinja's grammar natively. Malformed Jinja that the lexer can't tokenize is returned unchanged, preserving the existing graceful-degradation contract with `parse_sql`. Five new tests cover the previously-broken edge cases and the error-handling paths.

### File Changes
| `src/dbt_bouncer/sql_utils.py` | Modified | Replaced regex-based Jinja stripping with jinja2 lexer token walk; added `_get_jinja_lexer_environment()` helper; added `lru_cache` to `neutralize_jinja` |
| `tests/unit/test_sql_utils.py` | Modified | Added 5 new test cases: delimiter-in-string, nested braces, whitespace control, malformed input, and unlexable input returned unchanged |

### Review Findings

<details>
<summary>All clear — expand for details</summary>

- ✅ **Python Quality**: Type hints on all function signatures, clean `TYPE_CHECKING` guard for jinja2, proper `lru_cache` usage (maxsize=2048 matches `parse_sql`), lazy imports inside the function body to defer jinja2's ~25ms import cost.
- ✅ **Check Structure**: N/A — this is a shared utility module, not a check.
- ✅ **Test Coverage**: Five new tests cover the three edge cases the regex couldn't handle (delimiter-in-string, nested braces, whitespace control) plus both error paths (malformed input returns a string, unlexable input returns the original unchanged). Existing tests for expression placeholders, comment stripping, statement dropping, and leading config are preserved.
- ✅ **Security**: `autoescape=True` is set on the Jinja environment with a clear comment explaining it only affects rendering (not lexing). No unsafe deserialization, no secrets, no command injection vectors.
- ✅ **Style**: Well-documented with clear docstrings (Parameters/Returns sections). Snake_case naming throughout. No dead code — the three private regex patterns (`_JINJA_STATEMENT_PATTERN`, `_JINJA_CONFIG_PATTERN`, `_JINJA_EXPRESSION_PATTERN`) were removed and confirmed unused elsewhere via grep.
- ✅ **Breaking Changes**: Public API is preserved — `neutralize_jinja`, `parse_sql`, and `JINJA_COMMENT_PATTERN` all keep their existing signatures and semantics. The removed private regexes had no external consumers.
- ✅ **Documentation**: Docstrings updated to describe the lexer-based approach, including the graceful-degradation contract for malformed input. Module-level comment on `JINJA_COMMENT_PATTERN` updated to explain why it remains public.
- ✅ **Cross-file version consistency**: No version bumps in this PR.

</details>

---

pgoslatara (approved):

PR #984, opened 2026-07-21
perf: execute checks sequentially instead of via ThreadPoolExecutor

pgoslatara:
Checks operate on already-parsed, in-memory dbt artifacts and are CPU-bound pure-Python work (regex matching, proxy attribute access, string ops) that never releases the GIL. The `ThreadPoolExecutor` in `Executor.run` therefore could never run them in parallel — it only added thread-scheduling and GIL-contention overhead, plus large tail-latency spikes on big projects. This replaces the threaded, class-batched execution (added in #579, batched in #621) with a straight sequential loop, which is both faster and far more predictable.

## Evidence

Profiled with the `tests/benchmark` synthetic manifest (32 checks). The runner is ~94% of wall time and the execute phase dominates it. Comparing the original threaded path against sequential, back-to-back in the same process (so machine load hits both equally):

| Scale | Original (threaded + batching) | Sequential | Improvement |
|---|---|---|---|
| 1,000 models (17,636 check runs) | ~245 ms | ~180 ms | ~27% |
| 5,000 models (88,134 check runs) | ~1.22 s | ~0.89 s | ~27% |

Isolated micro-benchmarks of the execute phase alone showed the threaded version is 20–28% slower at every scale, and — because thread scheduling and GIL contention are non-deterministic — it also produced multi-second worst-case spikes (max 3.6 s on one full run) that the sequential path eliminates. The gain holds and widens with scale, confirming it is the GIL rather than one-off thread-startup cost.

## Changes

- `Executor.run` now iterates checks sequentially.
- Removed the now-unused batching machinery (`_execute_batch`, `_MAX_BATCH_SIZE`) and the `threading` / `concurrent.futures` imports.
- The progress bar is retained but refreshed ~100 times total instead of once per check, so it adds negligible overhead at scale.

`Executor.run`'s public contract (result dict shape, outcomes, severity handling) is unchanged; the existing executor and runner unit tests pass without modification, as does the integration suite.

## Trade-off

Removing the thread pool means custom user checks that do blocking I/O would no longer overlap. That is not a pattern the built-in checks use (all data is parsed into memory up front), so the default is the right one. If demand for it appears, a parallel path could be reintroduced as an explicit opt-in flag rather than the default.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -48,7 +48,7 @@ make install
   - `exceptions.py` — `DbtBouncerFailedCheckError` and `NestedDict`
   - `patterns.py` — abstract check pattern ABCs
 - `src/dbt_bouncer/runner.py` — orchestrates check execution
-- `src/dbt_bouncer/executor.py` — parallel execution with `ThreadPoolExecutor`
+- `src/dbt_bouncer/executor.py` — sequential check execution with progress tracking
 - `tests/` — mirrors `src/` structure; fixtures in `tests/fixtures/`
 
 ### Check System
@@ -59,7 +59,7 @@ Checks are written using the `@check` decorator (preferred) or as class-based `B
 
 1. `runner.py` iterates checks, determines the resource type from the function's first positional parameter (decorator API) or class annotations (class-based API)
 2. Injects the per-iteration resource into the check
-3. `Executor` batches checks by class and runs them via `ThreadPoolExecutor`
+3. `Executor` runs checks sequentially with a Rich progress bar
 4. Failed checks call `fail()` (decorator API) or raise `DbtBouncerFailedCheckError` (class-based API)
 
 **Resource types:** catalog_node, catalog_source, exposure, macro, model, run_result, seed, semantic_model, snapshot, source, test, unit_test
--- a/src/dbt_bouncer/executor.py
+++ b/src/dbt_bouncer/executor.py
@@ -1,12 +1,9 @@
-"""Check execution engine with batching and parallel execution."""
+"""Check execution engine with sequential execution and progress tracking."""
 
 from __future__ import annotations
 
 import logging
-import threading
 import traceback
-from collections import defaultdict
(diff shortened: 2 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Replace ThreadPoolExecutor with sequential check execution to eliminate GIL-contention overhead and tail-latency spikes.**

**Type:** `perf`

### Summary
dbt-bouncer checks are CPU-bound pure-Python work that never releases the GIL, so the existing `ThreadPoolExecutor` could not actually parallelise them — it only added thread-scheduling overhead and unpredictable latency spikes on large projects. This PR removes the threading machinery (batching by check class, `ThreadPoolExecutor`, `progress_lock`) and replaces it with a simple sequential loop that throttles Rich progress-bar updates to ~100 refreshes total. `AGENTS.md` is updated to reflect the new execution model.

### File Changes
| `src/dbt_bouncer/executor.py` | Modified | Removed `ThreadPoolExecutor`, batching logic, `_execute_batch`, threading imports, and `_MAX_BATCH_SIZE`; replaced with sequential loop and throttled progress updates |
| `AGENTS.md` | Modified | Updated architecture docs to reflect sequential execution instead of parallel `ThreadPoolExecutor` |

### Review Findings
- ✅ No issues with **Python quality** — dead imports (`threading`, `defaultdict`, `concurrent.futures`) cleanly removed; remaining code is well-typed and minimal.
- ✅ No issues with **check structure** — executor only calls `check.execute()`, no change to the check API.
- ✅ No issues with **test coverage** — this is infrastructure-only; the comment references the benchmark suite (`tests/benchmark`) for validation. No new public API surface was added.
- ✅ No issues with **style** — clean formatting, good explanatory comments justifying the change.
- ✅ No issues with **breaking changes** — `Executor.run()` signature and return type are unchanged; `_execute_batch` was private API.
- ✅ No issues with **documentation** — `AGENTS.md` updated in the same PR to match the new architecture.
---

mattcristal:
Nice!

PR #985, opened 2026-07-21
perf: bind resource in executor instead of per-resource model_copy

pgoslatara:
Now that checks execute sequentially (#984), the runner no longer needs to `model_copy()` a fresh check instance for every matched resource. That shallow copy — ~17.6k of them on a 1,000-model project, more on larger ones — was there so each concurrently-executing check could carry its own bound resource. With sequential execution a single check-config instance can serve all of its resources in turn: the assembly step now stores the resource alongside the check, and the executor binds it onto the shared instance immediately before calling `execute()`.

This is safe precisely because execution is sequential — reusing one instance across resources would be a data race under the old thread pool. `report_dry_run` only reads each check's class and count (never per-instance resource state), so it is unaffected.

## Evidence

Measured on the `tests/benchmark` synthetic manifest (32 checks), comparing this branch against `main`:

| Benchmark | main | this branch | Reduction |
|---|---|---|---|
| `test_runner_match` (match phase) | 271 ms | 73 ms | ~73% |
| `test_runner` (match + execute + report) | 442 ms | 232 ms | ~46% |

The "Match & copy" phase drops from ~21% to ~12% of end-to-end runtime; the effect grows with project size (match phase ~1.54 s → ~0.64 s at 5,000 models). `model_copy` is genuinely the dominant cost of the match phase — eliminating it, rather than making it cheaper, is what recovers the time.

## Changes

- `runner.py`: `_assemble_checks_to_run` no longer copies per match. It sets the check context once per config and appends the shared `check` plus the `resource`/`iterate_value` needed to bind it later. `CheckToRun` gains `resource` and `iterate_value` fields.
- `executor.py`: `_execute_check` binds `resource` onto the check instance immediately before `execute()`. Context-only checks (no resource) are executed as-is, so the existing executor tests — which pass dicts without a `resource` key — are unaffected.

All 1,117 unit tests and 22 integration tests pass unchanged; `Executor.run`'s public contract (result dict shape, outcomes, severity handling) is unchanged.

### DIFF
--- a/src/dbt_bouncer/executor.py
+++ b/src/dbt_bouncer/executor.py
@@ -29,6 +29,14 @@ class Executor:
 
         """
         logging.debug(f"Running {check['check_run_id']}...")
+        # Bind this run's resource onto the shared check instance immediately
+        # before executing. Assembly stores the resource alongside the check
+        # rather than pre-copying an instance per resource; sequential execution
+        # makes reusing one instance across its resources safe. Context-only
+        # checks have no resource and are executed as-is.
+        resource = check.get("resource")
+        if resource is not None:
+            check["check"].set_resource(resource, check["iterate_value"])
         try:
             check["check"].execute()
             check["outcome"] = CheckOutcome.SUCCESS
--- a/src/dbt_bouncer/runner.py
+++ b/src/dbt_bouncer/runner.py
@@ -76,13 +76,21 @@ def _build_check_run_id(check: Any, resource: Any, iterate_value: str) -> str:
 
 
 class CheckToRun(TypedDict):
-    """A single check instance ready for execution, with its run context."""
+    """A single check ready for execution, with its run context.
+
+    ``check`` is a shared check-config instance; ``resource`` and ``iterate_value``
+    tell the executor which resource to bind onto it immediately before calling
+    ``execute()``. Because checks run sequentially, one instance can serve all of
+    its resources in turn — no per-resource copy is needed.
+    """
 
     check: Any
     check_run_id: str
     failure_message: NotRequired[str]
     file_path: NotRequired[str | None]
+    iterate_value: NotRequired[str | None]
     outcome: NotRequired[str]
+    resource: NotRequired[Any]
     severity: str
     unique_id: NotRequired[str | None]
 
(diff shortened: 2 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Eliminates per-resource `model_copy()` calls by reusing a single shared check instance across all its matched resources, binding the resource in the executor immediately before execution.**

**Type:** `refactor`

### Summary
Previously, `_assemble_checks_to_run` created a shallow Pydantic `model_copy()` of each check for every matched resource, then called `set_resource` and `set_context` on each copy upfront. This PR removes those per-resource copies entirely: the shared check-config instance is stored once in the run list alongside the resource and iterate-value, and the executor binds the resource onto the shared instance immediately before calling `execute()`. This is safe because checks now run sequentially (established in a prior PR), so no two executions share the instance concurrently. The change avoids ~17k+ `model_copy` calls on typical projects.

The two files work together: `runner.py` stops copying and instead stores `resource`/`iterate_value` in the `CheckToRun` dict, while `executor.py` reads those fields and calls `set_resource` just-in-time before `execute()`.

### File Changes
| `src/dbt_bouncer/runner.py` | Modified | Removes per-resource `model_copy()` and `set_resource` calls during assembly; stores `resource` and `iterate_value` in the `CheckToRun` TypedDict; moves `set_context` to once per check; updates docstrings |
| `src/dbt_bouncer/executor.py` | Modified | Adds just-in-time `set_resource` call on the shared check instance before `execute()`, guarded by a `resource is not None` check for context-only checks |

### Review Findings
- ✅ No issues with **Python quality** — type annotations on `CheckToRun` are correct (`NotRequired` for optional fields), no dead code, clean use of TypedDict.
- ✅ No issues with **check structure** — `set_resource`/`set_context`/`execute()` call sequence is preserved; context-only checks (no `iterate_over_value`) still work correctly since the executor guards with `resource is not None`.
- ✅ No issues with **test coverage** — this is a performance refactor of internal plumbing; the existing check and benchmark test suites cover the behavior. No new public API surface was added.
- ✅ No issues with **security** — no deserialization, no user input handling, no secrets.
- ✅ No issues with **style** — comments are clear and explain the *why*; `CheckToRun` fields are alphabetically ordered; code is consistent with surrounding style.
- ✅ No issues with **breaking changes** — `CheckToRun` is an internal TypedDict (not part of the public API); the new fields are `NotRequired` so any existing code constructing these dicts is unaffected.
- ✅ No issues with **documentation** — docstrings on `CheckToRun` and `_assemble_checks_to_run` are updated to accurately reflect the no-copy behaviour.
---

PR #992, opened 2026-07-23
feat: add check_model_does_not_rejoin_upstream_concepts

pgoslatara:
Adds `check_model_does_not_rejoin_upstream_concepts`, which fails when a model reads from both a parent `B` and one of `B`'s own parents `A`. When `B` exists only to feed this model, `A`'s columns arrive by two different routes and the two paths can drift apart as the logic evolves.

The check is evaluated from the child's perspective, which fits dbt-bouncer's per-model iteration: for each parent `B` of the model under test, if `B`'s parents intersect the model's own parents and `B` has exactly one downstream model, the model is flagged. The single-consumer condition is what separates a genuine rejoin from a legitimately shared concept — a `B` with several consumers is a reusable intermediate, not an accidental detour.

Shared ancestors may be sources as well as models, so the source-side variant of the pattern is caught too.

Closes a gap identified by comparing dbt-bouncer against [dbt-lint-yaml's rule set](https://github.com/VDFaller/dbt-lint-yaml/blob/main/docs/rules/index.md) (`rejoining_of_upstream_concepts`, originally from dbt-project-evaluator). Nothing in dbt-bouncer currently inspects the DAG for this shape.

Note: `schema.json` is generated — if this conflicts with another check-adding PR, resolve by re-running `mise run generate-schema` rather than merging by hand.

Draft while the batch of related gap-closing checks is reviewed together.

### On the rule code gap

This PR uses a non-contiguous code, which looks like an oversight when the PR is read on its own. It is deliberate: this is one of a batch of gap-closing checks opened together, and the codes were allocated across the whole batch up front so that two PRs could not claim the same number.

| PR | Check | Code |
|---|---|---|
| #991 | `check_model_does_not_directly_join_to_source` | `MO048` |
| #992 | `check_model_does_not_rejoin_upstream_concepts` | `MO049` |
| #993 | `check_model_min_downstream_models` | `MO050` |
| #995 | `check_model_has_properties_file` | `MO051` |
| #994 | `check_source_top_level_description_populated` | `SO017` |

If any of the batch is rejected, the survivors leave gaps. Gaps are harmless — codes are identifiers, not indices — and `AGENTS.md` forbids renumbering published codes. But nothing here is published until merged, so if you want a contiguous range, renumber before the first merge rather than after.

PR #991, opened 2026-07-23
feat: add check_model_does_not_directly_join_to_source

pgoslatara:
Adds `check_model_does_not_directly_join_to_source`, which fails when a model's upstream dependencies include both a source and a model. Joining raw source data straight onto an already-transformed model bypasses the staging layer, so the renaming, casting and cleaning that every other consumer of that source gets is silently skipped.

Only parents with a `model` resource type count as the curated side of the join, matching dbt-project-evaluator's `fct_direct_join_to_source` exactly. Seed and snapshot parents are deliberately not counted — extending that is a follow-up rather than a silent divergence from the rule everyone already knows.

Closes a gap identified by comparing dbt-bouncer against [dbt-lint-yaml's rule set](https://github.com/VDFaller/dbt-lint-yaml/blob/main/docs/rules/index.md) (`direct_join_to_source`). The existing `check_lineage_source_cannot_be_used` only bans sources by path regex and `check_model_depends_on_multiple_sources` only counts sources, so neither expresses the mixed-parent condition.

Draft while the batch of related gap-closing checks is reviewed together.

PR #993, opened 2026-07-23
feat: add check_model_min_downstream_models

pgoslatara:
Adds `check_model_min_downstream_models`, the lower-bound counterpart to the existing `check_model_max_fanout`. A model that nothing downstream references still has to be built, tested and understood on every run, but nothing consumes its output — usually the residue of a retired mart or a refactor that stranded an intermediate model.

This mirrors the existing `check_source_min_downstream_models`, so sources and models now have symmetric coverage.

Downstream models **and snapshots** both count as consumers. `ctx.children_by_unique_id` is derived from `ctx.models` only, so snapshot consumers are counted separately — without that, a staging model feeding only a snapshot reads as dead. The test project has exactly such a model (`stg_raw_orders_for_snapshot`, consumed by two snapshots), which is what surfaced the case. Tests and unit tests deliberately do not count, since they do not make a model useful to anyone.

Running this against the example project also flagged four genuinely unreferenced intermediate models that exist only as test fixtures, so the example config scopes the check to `^models/staging`.

Closes a gap identified by comparing dbt-bouncer against [dbt-lint-yaml's rule set](https://github.com/VDFaller/dbt-lint-yaml/blob/main/docs/rules/index.md) (`dead_model`). `check_model_has_exposure` is related but stricter and different — it requires an exposure rather than any downstream consumer, and remains the right check for the final layer.

Note: `schema.json` is generated — if this conflicts with another check-adding PR, resolve by re-running `mise run generate-schema` rather than merging by hand.

Draft while the batch of related gap-closing checks is reviewed together.

### On the rule code gap

This PR uses a non-contiguous code, which looks like an oversight when the PR is read on its own. It is deliberate: this is one of a batch of gap-closing checks opened together, and the codes were allocated across the whole batch up front so that two PRs could not claim the same number.

| PR | Check | Code |
|---|---|---|
| #991 | `check_model_does_not_directly_join_to_source` | `MO048` |
| #992 | `check_model_does_not_rejoin_upstream_concepts` | `MO049` |
| #993 | `check_model_min_downstream_models` | `MO050` |
| #995 | `check_model_has_properties_file` | `MO051` |
| #994 | `check_source_top_level_description_populated` | `SO017` |

If any of the batch is rejected, the survivors leave gaps. Gaps are harmless — codes are identifiers, not indices — and `AGENTS.md` forbids renumbering published codes. But nothing here is published until merged, so if you want a contiguous range, renumber before the first merge rather than after.


pgoslatara:
## 🔍 AI PR Review
> **Adds a new `check_model_min_downstream_models` check (MO050) that flags models with too few downstream consumers, helping teams detect dead or orphaned models.**

**Type:** `feature`

### Summary
This PR introduces a new lineage check, `check_model_min_downstream_models`, that ensures models are referenced by at least a configurable number of downstream models and/or snapshots. This is the counterpart to the existing `check_model_max_fanout` — where max-fanout caps the upper bound, this new check enforces a lower bound. The check counts both model children (via `ctx.children_by_unique_id`) and snapshot consumers (by scanning `ctx.snapshots` depends_on), so a model feeding only a snapshot is not falsely flagged as dead. All six files form a coherent unit: the check implementation, its rule code enum entry, JSON schema, example config, documentation, and tests.

<details>
<summary>📄 File Changes (6 files)</summary>

| `src/dbt_bouncer/checks/manifest/models/lineage.py` | Modified | Adds the `check_model_min_downstream_models` function with `@check(code="MO050")` |
| `src/dbt_bouncer/enums.py` | Modified | Adds `CHECK_MODEL_MIN_DOWNSTREAM_MODELS = "MO050"` to the `ModelRuleCode` enum |
| `schema.json` | Modified | Adds the `CheckModelMinDownstreamModels` schema definition and references |
| `docs/checks/rule_codes.md` | Modified | Adds MO050 to the rule codes table and bumps the MO count to 50 |
| `dbt-bouncer-example.yml` | Modified | Adds an example entry for the new check scoped to `^models/staging` |
| `tests/unit/checks/manifest/models/test_lineage.py` | Modified | Adds `TestCheckModelMinDownstreamModels` (6 cases) and `TestCheckModelMinDownstreamModelsInvalidParam` (1 case) |

</details>

### Review Findings
- ✅ **Python Quality**: Clean imports (reuses existing `Annotated`, `Field`, `check`, `fail`, `get_clean_model_name`), proper type annotation on the parameter via `Annotated[int, Field(gt=0)]`, no dead code.
- ✅ **Check Structure**: Correct use of `@check(code="MO050")` decorator, proper `(model, ctx, *, param)` parameter ordering, `fail()` used for check failure, rule code added to `ModelRuleCode` enum in alphabetical order.
- ✅ **Test Coverage**: Happy paths cover single downstream model, snapshot-only consumer, higher minimum with two models, and mixed model+snapshot consumers. Unhappy paths cover dead model (no matching downstream) and below-higher-minimum. Invalid parameter test confirms `min_number_of_models=0` is rejected via Pydantic validation. Good coverage.
- ✅ **Security**: No deserialization, no secrets, no command injection risk. Pure data inspection.
- ✅ **Style**: snake_case naming, enum entries in alphabetical order, schema entries in alphabetical order, consistent with existing check patterns.
- ✅ **Breaking Changes**: No breaking changes — purely additive new check, new rule code, new schema entry.
- ✅ **Documentation**: Comprehensive docstring with Rationale, Parameters, Receives, Other Parameters, and Example(s) sections. Rule codes doc updated. Example config updated.
- ✅ **Cross-file version consistency**: No version-pinned tools changed in this PR.

---

PR #989, opened 2026-07-22
feat: introduce number coding for checks

mattcristal:
## What was done

This PR is a proposal implementation for: https://github.com/godatadriven/dbt-bouncer/issues/733

## Implementation

Now, all the checks have an **assigned code** to them, which is unique.

Using a **2-letter prefix + 3-digit number** (e.g. `MO001`) provides clean sorting, alignment, and plenty of capacity (up to 999 checks per category):

Prefix | Category | Examples
-- | -- | --
MO | Model Manifest Checks | MO001 (check_model_has_description), MO002 (check_model_names)
MA | Macro Checks | MA001 (check_macro_has_description)
SO | Source Checks | SO001 (check_source_has_description), SO002 (check_source_name)
SE | Seed Checks | SE001 (check_seed_has_description)
EX | Exposure Checks | EX001 (check_exposure_based_on_model)
SN | Snapshot Checks | SN001 (check_snapshot_has_description)
TE | Test Checks | TE001 (check_test_name)
UT | Unit Test Checks | UT001 (check_unit_test_has_description)
SM | Semantic Model Checks | SM001 (check_semantic_model_has_description)
CA | Catalog Checks | CA001 (check_catalog_columns_have_types)
RR | Run Results Checks | RR001 (check_run_results_max_execution_time)

Implementation example:

```python
class MetadataRuleCode(StrEnum):
    """Rule codes for metadata checks."""

    CHECK_PROJECT_NAME = "ME001"
```

## dbt-bouncer list

Now generates a nicer looking rich table:

Screenshot 2026-07-22 at 18 19 37

But now this is also possible:

```shell
# Filter to catalog checks only
uv run dbt-bouncer list --group catalog

# Filter to manifest checks only
uv run dbt-bouncer list -g manifest_checks

# Output catalog checks as JSON
uv run dbt-bouncer list -g catalog_checks --output-format json
```

## More info

- Dual Referencing: dbt-bouncer.yml configurations accept either name: check_model_access or name: MO001 or code: MO001.
- dbt-bouncer list Display: Outputs checks formatted with rule codes (e.g. [MO001] check_model_access).
- skip_checks Support: Model metadata skip_checks accepts both check names and rule codes.
Full Backward Compatibility: All existing configurations and tests work unchanged.

PR #997, opened 2026-07-23
docs: use an abstract placeholder for the rule code example

pgoslatara:
Two small corrections to the rule-code guidance added in #989.

**The template example used a real code.** `AGENTS.md` showed `@check(code="MO048")` as the scaffold to copy. `MO048` was the next free model code when #989 was written, but it is a real, assignable code — and #991 now claims it. A contributor copying the template verbatim would either collide with a live code or silently ship the placeholder. Changed to `XX000`, which matches the abstract form the prose two paragraphs below already uses (`pass it as @check(code="XX000")`), so the file is now self-consistent and the example can never go stale as codes are assigned.

**A leftover bullet contradicted the new requirement.** Directly under "Every check needs a unique rule code — pass it as `@check(code="XX000")`" sat "`@check` with no arguments — name and `iterate_over` are inferred from the function name", written before #989 and not updated. Reworded to say `code` is the only argument `@check` takes, keeping the point about name and `iterate_over` inference.

Docs only — no source, schema, or generated pages affected.

Draft, as with the rest of the batch.

**The `/new-check` skill had the same problem, worse.** `.agents/skills/new-check/SKILL.md` (reached via the tracked `.claude/skills` symlink) still stated "`@check` is a bare decorator — it takes **no arguments**" and none of its four code templates passed a `code` at all. Anything scaffolded from it produced a codeless check that fails the `rule-codes-doc-check` hook. Updated:

- All four templates now pass a code. The three that illustrate real checks use those checks' actual codes (`MO021`, `MO038`, `MO044`); the generic scaffold uses `XX000`.
- New "Assign a rule code" section covering both required steps (decorator argument **and** the `*RuleCode` enum member in `enums.py`), the next-free-number rule, the no-renumbering rule, and the full 13-prefix table.
- The verify step now runs `mise run generate-rule-codes-doc` alongside `generate-schema`, and notes what `rule-codes-doc-check` rejects.

### DIFF
--- a/.agents/skills/new-check/SKILL.md
+++ b/.agents/skills/new-check/SKILL.md
@@ -15,12 +15,12 @@ Follow these steps to add a new check to dbt-bouncer.
 
 ## 2. Write the Check
 
-Use the `@check` decorator (bare, no arguments). Everything is inferred from the function signature:
+Use the `@check` decorator, passing the rule code. Everything else is inferred from the function signature:
 
 ```python
 from dbt_bouncer.check_framework.decorator import check, fail
 
-@check
+@check(code="XX000")
 def check_model_xxx(model):
     """Check description."""
     if some_condition:
@@ -29,8 +29,9 @@ def check_model_xxx(model):
 
 ## 3. Decorator API Reference
 
-`@check` is a bare decorator — it takes **no arguments**. All metadata is inferred from the function signature:
+`code` is the only argument `@check` takes. All other metadata is inferred from the function signature:
 
+- **code** — the check's unique rule code, e.g. `MO048`. See "Assign a rule code" below.
 - **name** — the function name (must match the `name:` value in YAML config).
 - **iterate_over** — the first positional parameter (excluding `ctx`). If there are none, the check is global (runs once with context only).
 - **params** — keyword-only arguments (after `*`) become user-configurable Pydantic fields.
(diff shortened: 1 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Updates documentation to reflect the mandatory `code` argument for the `@check` decorator, replacing the outdated "bare decorator" guidance.**

**Type:** `docs`

### Summary
Following the introduction of number coding for checks (#989), the `@check` decorator now requires a `code` argument. This PR updates both `AGENTS.md` and the `new-check` skill documentation to accurately describe the new API: `@check(code="XX000")` instead of the previous bare `@check`. It also adds a new "Assign a rule code" section to the skill with the prefix reference table, enum registration instructions, and the `generate-rule-codes-doc` mise task.

The two files serve different audiences but convey the same information: `AGENTS.md` is the shared project-level contributor guide, while `.agents/skills/new-check/SKILL.md` is the step-by-step scaffolding skill used by AI agents.

### File Changes
| `AGENTS.md` | Modified | Replaced the example placeholder code `MO048` with `XX000` in the template, and clarified that `code` is the only argument `@check` takes |
| `.agents/skills/new-check/SKILL.md` | Modified | Updated all `@check` examples to include a `code` argument, added "Assign a rule code" section with prefix table and enum instructions, added `generate-rule-codes-doc` to the validation commands |

<details>
<summary>Review Findings</summary>

</details>

---

PR #995, opened 2026-07-23
feat: add check_model_has_properties_file

pgoslatara:
Adds `check_model_has_properties_file`, which fails when a model has no `patch_path`, i.e. it is not declared in any `.yml` properties file. Without one there is nowhere to hang a description, column definitions, tests, contracts or meta keys, so every other documentation and testing check silently has nothing to inspect.

`check_model_property_file_location` already reports undocumented models, but only as a side effect of enforcing dbt Labs' `_<directory>__models.yml` naming convention — a project that wants to require a properties file without adopting that convention currently has no way to say so. The docstring cross-references both directions so the overlap is explicit.

Closes a gap identified by comparing dbt-bouncer against [dbt-lint-yaml's rule set](https://github.com/VDFaller/dbt-lint-yaml/blob/main/docs/rules/index.md) (`missing_properties_file`). Of the six gap-closing PRs in this batch, this is the smallest and the most redundant with what already exists — reasonable to reject on those grounds if you would rather not add a near-duplicate.

Note: `schema.json` is generated — if this conflicts with another check-adding PR, resolve by re-running `mise run generate-schema` rather than merging by hand.

Draft while the batch of related gap-closing checks is reviewed together.

All four codes in this batch are now published and contiguous: `MO048` (#991), `MO049` (#992), `MO050` (#993) and `MO051` here.

PR #998, opened 2026-07-23
fix: disable rich emoji substitution in the console report

pgoslatara:
Builds the console report with `emoji=False` so rich stops substituting emoji shortcodes into check run IDs and failure messages.

`check_run_id` is built as `<check_name>:<index>:<resource>`. The 100th check of a run therefore contains `:100:`, which rich reads as the shortcode for 💯 and replaces. The reported ID becomes `check_model_names💯model.my_model` — the check index is silently destroyed, and the ID no longer matches anything a user can search for.

On Windows it is worse than cosmetic: the substituted character is unencodable in the cp1252 console, so printing the report raises `UnicodeEncodeError: 'charmap' codec can't encode character '\U0001f4af'` and the run aborts with exit code 1 after all checks have already passed.

`escape()` from `rich.markup` (added in #975 for the square-bracket case) does not help here — it guards markup tags, not emoji shortcodes. Verified directly:

```python
Console(emoji=True ).print(escape("check_x:100:my_model"))  # -> check_x💯my_model
Console(emoji=False).print(escape("check_x:100:my_model"))  # -> check_x:100:my_model
```

`:100:` is not the only exposure — any `:<shortcode>:` sequence in a failure message or resource name is substituted too. Disabling emoji on the console covers all of them. Both consoles in `reporter.py` are changed: the failure table and the dry-run table.

The trigger is a run producing 100 or more checks, so it affects any real-world config rather than an edge case. It is latent on `main` today and is currently blocking #992, #993 and #995, each of which pushes the example project's check count past 100 on Windows CI.

The regression test asserts `:100:` survives and 💯 does not appear; it fails on `main` and passes with this change.

### DIFF
--- a/src/dbt_bouncer/reporting/reporter.py
+++ b/src/dbt_bouncer/reporting/reporter.py
@@ -81,7 +81,7 @@ class Reporter:
             )
             counts[check_name, resource_type] += 1
 
-        console = Console()
+        console = Console(emoji=False)
         table = Table(
             title="[bold cyan]Dry run — checks that would execute[/bold cyan]",
             title_justify="left",
@@ -129,7 +129,7 @@ class Reporter:
             else:
                 num_checks_success += 1
 
-        console = Console()
+        console = Console(emoji=False)
 
         if num_checks_error > 0 or num_checks_warn > 0:
             logger = logging.error if num_checks_error > 0 else logging.warning
--- a/tests/unit/reporting/test_reporter.py
+++ b/tests/unit/reporting/test_reporter.py
@@ -139,6 +139,37 @@ def test_report_results_escapes_rich_markup_in_console(capsys, monkeypatch):
     assert "[a-z0-9]" in out
 
 
+def test_report_results_does_not_substitute_emoji_shortcodes(capsys, monkeypatch):
+    """Check run IDs containing `:<digits>:` are not turned into emoji.
+
+    `check_run_id` is built as `<check_name>:<index>:<resource>`, so the 100th
+    check of a run yields `...:100:...` — which rich reads as the shortcode for
+    💯 and substitutes. `rich.markup.escape` does not prevent this (it only
+    guards square-bracket tags), so the console is built with `emoji=False`.
+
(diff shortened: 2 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Disables Rich emoji substitution in console output to prevent `:100:` in check run IDs from being rendered as emoji, which also caused `UnicodeEncodeError` on Windows.**

**Type:** `bugfix`

### Summary
When dbt-bouncer runs 100+ checks, the `check_run_id` format (`<check_name>:<index>:<resource>`) produces strings like `:100:`, which Rich interprets as the shortcode for the 💯 emoji. This mangled the displayed ID and, on Windows with cp1252 consoles, raised a `UnicodeEncodeError` that aborted the run entirely. The fix passes `emoji=False` to both `Console()` instantiations in the reporter, since `rich.markup.escape()` only guards square-bracket tags, not emoji shortcodes. A targeted regression test is added to verify the fix.

### File Changes
| `src/dbt_bouncer/reporting/reporter.py` | Modified | Pass `emoji=False` to both `Console()` calls (dry-run and results reporting) to disable emoji shortcode substitution |
| `tests/unit/reporting/test_reporter.py` | Modified | Add regression test asserting `:100:` is preserved literally and the 💯 character does not appear in output |

### Review Findings
- ✅ No issues with **Python quality** — no new function signatures, clean minimal change, no dead code.
- ✅ No issues with **check structure** — changes are in the reporter, not in check definitions.
- ✅ No issues with **test coverage** — new test covers the exact scenario (`:100:` shortcode) with both a positive assertion (`:100:` preserved) and a negative assertion (no 💯 emoji). The docstring thoroughly explains the root cause.
- ✅ No issues with **style** — change is minimal and consistent with existing code patterns.
- ✅ No issues with **breaking changes** — `emoji=False` only suppresses Rich's shortcode-to-emoji substitution; it does not affect markup formatting or any public API.
- ✅ No issues with **documentation** — no public API changes; test docstring serves as documentation.

---

PR #1006, opened 2026-07-29
fix: correct check_model_version_pinned_in_ref and expand versioning …

mattcristal:
## What was done

Expanding the unit tests for the **model versioning** checks (9 -> 37 tests, 100% line coverage of the module) uncovered two bugs in `check_model_version_pinned_in_ref`, both fixed here.

1. The check never fired for genuinely versioned models. It compared a downstream `ref`'s `name` against `model.unique_id.split(".")[-1]`. For a versioned model dbt emits `unique_id` as `model.<package>.<name>.v<version>`, so that expression yields the version (`v2`) rather than the model name (`customers`), and the comparison never matched. It now compares against `model.name`, which dbt populates without the version suffix.

2. The check fired for every unversioned model. An unversioned model has no version for a downstream `ref` to pin, so its consumers were all reported as violations. The check now returns early when `model.version` is None.

Verified against tests/fixtures/dbt_111: before this change 21 of the project's unversioned models were incorrectly flagged, and the one genuinely unpinned versioned model was not flagged. After, both are correct.

Also:

- The failure message now uses `get_clean_model_name`, so it identifies which version of a versioned model is unpinned (`customers_v2`) rather than the ambiguous bare name shared by every version.
- Removed a stray `)` from the `check_model_version_allowed` failure message.
- Documented in the docstring that only versioned models are checked, and that only downstream models are considered, so a `ref` from a test or exposure does not trigger the check.


mattcristal:
moved back to draft

mattcristal:
@pgoslatara can you approve the workflow again, please?

pgoslatara:
## 🔍 AI PR Review
> **Fixes two bugs where a version of `0` was silently treated as "absent" and where `check_model_version_pinned_in_ref` was a no-op for every genuinely versioned model, plus rewrites the test suite for comprehensive coverage.**

**Type:** `bugfix`

### Summary
This PR fixes two truthiness-vs-identity bugs in the model versioning checks and a name-comparison bug that made MO047 (`check_model_version_pinned_in_ref`) silently pass for every versioned model. The old code used `if model.version` / `not getattr(ref, "version", None)` which treated a version of `0` as absent; both are now `is not None` checks. More critically, MO047 compared `ref.name` (e.g. `"customers"`) against `model.unique_id.split(".")[-1]` (e.g. `"v2"` for a versioned model), which could never match — the check now correctly compares against `model.name`. The test suite is rewritten from scratch with minimal fixtures, helper functions, and thorough parametrised coverage of edge cases including version-zero, unversioned models, non-model downstream nodes, and deduplication of violations.

### File Changes
| `src/dbt_bouncer/checks/manifest/models/versioning.py` | Modified | Fix truthiness bugs (`is not None` instead of truthiness), fix ref name comparison (`model.name` instead of `unique_id.split(".")[-1]`), add early return for unversioned models, deduplicate violations with `any()`+`append`, use `get_clean_model_name` in error message, fix typo in closing paren |
| `tests/unit/checks/manifest/models/test_versioning.py` | Modified | Complete rewrite: minimal fixtures via `_MODEL_V2` and `_downstream_model`/`_manifest` helpers; parametrised happy+unhappy paths covering version-zero, unversioned models, non-model downstream nodes, missing child_map entries, deduplication, mixed pinned/unpinned refs, and error message assertions |
| `schema.json` | Modified | Updated docstrings for `CheckModelVersionAllowed` and `CheckModelVersionPinnedInRef` to document that only versioned models are checked |

### Review Findings
- ✅ No issues with **Python quality** — `is not None` identity checks are correct, new import of `get_clean_model_name` is clean, no dead code.
- ✅ No issues with **check structure** — both checks correctly use `@check` decorator and `fail()`, early return for unversioned models is the right pattern.
- ✅ No issues with **test coverage** — excellent parametrised coverage of happy/unhappy paths, edge cases (version zero, unversioned models, non-model downstream nodes, absent child_map, deduplication), and error message assertions via `match=`.
- ✅ No issues with **security** — no deserialization changes, no user input handling changes.
- ✅ No issues with **style** — snake_case naming, clean helper functions, informative `pytest.param` IDs, well-placed comments explaining *why*.
- ✅ No issues with **breaking changes** — the behavioural changes are bug fixes (version 0 now works, MO047 now actually detects violations); the only user-visible difference is that MO047 will start catching real violations it was silently ignoring.
- ✅ No issues with **documentation** — docstrings updated in both code and `schema.json` to clarify that only versioned models are checked.
---

pgoslatara (approved):

PR #1007, opened 2026-07-30
perf: set the log level on the logger, not only on the handler

pgoslatara:
`configure_console_logging` pinned the root logger at `DEBUG` and left the handler to filter, so every suppressed `logging.debug()` call in the codebase still ran the full record path — `makeRecord`, `LogRecord.__init__` and `findCaller()`, which walks the stack and splits file paths per frame — before the handler dropped it. The executor logs one debug line per check, so a 1,000-model project built and discarded roughly 17,600 records per run.

Setting the level on the logger lets `isEnabledFor()` short-circuit first. The logger keeps the lowest level any remaining non-`StreamHandler` handler asked for, so an embedding application's own handler still sees what it configured. The per-check debug call in the executor is additionally guarded behind a flag resolved once per `run()`, which removes the f-string too.

Measured on a 1,000-model synthetic project (17,636 checks): the executor's framework floor, with every check body replaced by a no-op, drops from **78.4 ms to 16.0 ms** (4.45 → 0.91 µs per check), and the execute phase from **157 ms to 89 ms**.

Two tests asserted DEBUG diagnostics at default verbosity and now pass `-v`; they only passed before because of this bug.


pgoslatara:
## 🔍 AI PR Review
> **Performance: avoid building tens of thousands of discarded LogRecords when DEBUG logging is off.**

**Type:** `refactor`

### Summary
On large dbt projects, the executor fires one `logging.debug()` call per check (tens of thousands). Previously the root logger was pinned at `DEBUG` and filtering was done only at the handler level, meaning Python built a full `LogRecord` (including `findCaller()` stack walks) for every call before the handler discarded it. This PR fixes the problem at two layers: (1) `configure_console_logging` now sets the logger level to the effective level so `isEnabledFor()` short-circuits early, and (2) the executor guards its three debug-log sites behind `self._debug_enabled` to also skip f-string interpolation. Two tests that relied on capturing DEBUG output without requesting verbose mode are updated to pass `-v`.

<details>

| `src/dbt_bouncer/reporting/logger.py` | Modified | Set root logger level to effective level instead of unconditionally `DEBUG`; preserve levels for non-StreamHandler handlers |
| `src/dbt_bouncer/executor.py` | Modified | Add `_debug_enabled` class var, resolve once in `run()`, guard all three `logging.debug()` calls |
| `tests/integration/test_integration_main.py` | Modified | Add `-v` flag so the test still captures DEBUG-level log output |
| `tests/unit/test_main.py` | Modified | Add `-v` flag so the custom-checks-dir DEBUG assertion still passes |

</details>

### Review Findings

- 🟡 **Suggestion:** `src/dbt_bouncer/executor.py:27` — `_debug_enabled` is a **class variable** defaulting to `False`. It works correctly because `run()` sets it on `self` (creating an instance attribute that shadows the class var), but it reads as a shared mutable class-level flag. Consider declaring it in `__init__` instead (`self._debug_enabled = False`) or annotating it only for type-checking purposes to make the intent clearer. Minor readability nit, not a bug.

- 🟡 **Suggestion:** `src/dbt_bouncer/reporting/logger.py:91-96` — The handler-level preservation logic filters by `type(h) is not logging.StreamHandler`, which intentionally excludes `StreamHandler` but *includes* its subclasses (e.g. `FileHandler`). This is consistent with the handler-removal logic below (line 99-101), so the design is sound, but a brief inline comment noting the intentional `is not` (vs `isinstance`) distinction would help future readers who might conflate the two.

- ✅ No issues with **Python quality** — type hints present on signatures, clean imports, proper use of class attributes.
- ✅ No issues with **test coverage** — both affected tests updated to pass `-v`; existing logger tests (`test_logging_debug_cli`, `test_logging_debug_env_var`, `test_logging_info`) already cover the DEBUG-with-`-v`, DEBUG-with-env-var, and INFO-without-`-v` paths respectively, confirming the new logger-level logic works at all three verbosity tiers.
- ✅ No issues with **security**.
- ✅ No issues with **style** — consistent with existing codebase conventions.
- ✅ No issues with **breaking changes** — the logger change is internal; public API and config schema are unchanged.
- ✅ No issues with **documentation** — no architecture changes; AGENTS.md does not need updating.

---

mattcristal:
Nice, I tried to work on this but I never finished

mattcristal:
This implementation is also cleaner compared to what I tried, good work

PR #1008, opened 2026-07-30
perf: raise the cyclic GC thresholds for CLI runs

pgoslatara:
Parsing a dbt project allocates a very large number of small containers — one proxy per manifest node, plus one per attribute access during checks, about 122,000 per run on a 1,000-model project. Against CPython's default gen-0 threshold of 700 the collector runs thousands of times over a heap that only ever grows.

Raising the thresholds is worth roughly a fifth of total runtime, nearly all of it in the parse and match phases. Measured end to end on 1,000 models: **179 ms → 139 ms**. `gc.disable()` would give 134 ms, so this recovers most of the benefit while leaving cycle collection on for very large projects. `gc.freeze()` after parsing was tried and is worse than raising the thresholds alone.

Only the CLI tunes this, and only in `main_callback` — a library must not reconfigure the collector of the process that imports it, so embedders calling `run_bouncer()` keep their own settings. One consequence: the benchmark suite calls `run_bouncer` directly, so `mise run test-benchmark` does not exercise this and will under-report real CLI speed.

Unrelated but unavoidable: ruff 0.16 requires the three `# noqa: F401` comments in this file to become `# ruff: ignore[unused-import]`, so `prek` fails on the file without it.


pgoslatara:
## 🔍 AI PR Review
> **Raises CPython's cyclic GC thresholds during CLI runs to eliminate unnecessary garbage collection overhead on large dbt projects.**

**Type:** `refactor`

### Summary
Parsing dbt manifests allocates many small proxy objects, triggering thousands of gen-0 garbage collections under CPython's default threshold of 700. On a 1,000-model project, GC was measured at ~20% of total runtime. This PR adds a `_tune_gc_for_cli()` helper that raises gen-0 to 20,000 (with gen-1/gen-2 at 25) and calls it only from the CLI entrypoint — after `--version`/`--help` early exits but before any heavy work — so library embedders calling `run_bouncer()` are unaffected.

### File Changes
| `src/dbt_bouncer/main.py` | Modified | Adds `_tune_gc_for_cli()` function and calls it in `main_callback` after early-exit flags are handled |

### Review Findings
- ✅ No issues with **Python quality** — function is well-typed (returns `None`), import is local to avoid side-effects at module level, docstring is thorough.
- ✅ No issues with **check structure** — no checks are added or modified.
- ✅ No issues with **test coverage** — this is a performance tuning change to GC thresholds with no behavioural change to test. The function is intentionally thin (single stdlib call) and the placement after early exits is straightforward.
- ✅ No issues with **security** — only calls `gc.set_threshold`, no external input.
- ✅ No issues with **style** — snake_case naming, private function prefix, clean formatting.
- ✅ No issues with **breaking changes** — tuning is scoped to the CLI path only; library API is unchanged.
- ✅ No issues with **documentation** — clear docstring explains the rationale, trade-offs, and scoping decision.
- ✅ No issues with **cross-file version consistency** — no version strings changed.

---

PR #1009, opened 2026-07-30
perf: compute per-resource match facts once per resource

pgoslatara:
Five values used while matching checks to resources depend only on the resource, but were recomputed for every (check, resource) pair: the run-ID suffix (an f-string plus `split`/`join`, 16 ms per run on its own), the file path, the `unique_id`, the cleaned path, and the `skip_checks` meta lookup. They now live in a `_ResourceFacts` record cached per resource.

Include/exclude filtering had the same shape — the regexes ran per (check, resource) even though most checks set neither pattern, and those that do share them. `_path_filtered_for` memoises the filtered list per `(iterate_value, include, exclude)`, with a fast path when both are unset.

The remaining per-pair work — the materialization filter and the resource's `skip_checks` — is in `_check_applies_to_resource`, called from the loop. Hoisting the path half out left `_should_run_check` and `_build_check_run_id` dead in production while their tests kept passing, so the live logic was untested; both are deleted and their tests now target the real predicate. Extracting it costs nothing (match 63.8 ms vs 63.9 ms inlined, back-to-back). This also adds coverage that was previously missing entirely: materialization filtering and skip-by-rule-code were only ever exercised through the dead function.

Measured on a 1,000-model synthetic project: match phase **40 ms → 27 ms (−31%)**. Check output is unchanged — verified byte-for-byte against the repo's own `dbt-bouncer-example.yml` (724 result entries).


pgoslatara:
## 🔍 AI PR Review
> **Performance: hoist per-resource computations out of the inner check loop and memoize path filtering by include/exclude pattern pairs.**

**Type:** `refactor`

### Summary
The hot loop in `_assemble_checks_to_run` previously recomputed string work (run-ID suffix, file path, unique_id, cleaned path) and path-matching regexes for every `(check, resource)` pair. This PR extracts those into a `_ResourceFacts` dataclass computed once per resource, and adds a second memoization layer (`_path_filtered_for`) that caches the filtered resource list per `(iterate_value, include, exclude)` tuple. Since most checks share the same include/exclude patterns (commonly both `None`), this avoids redundant regex evaluations across checks. The `_should_run_check` function is narrowed to `_check_applies_to_resource`, which now handles only the non-path predicates (materialization + skip_checks), with path filtering handled upstream. Tests are updated to match the new API surface.

### File Changes
| `src/dbt_bouncer/runner.py` | Modified | Introduces `_ResourceFacts` dataclass, `_pattern_key` helper, `_path_filtered_for` memoization cache; refactors `_build_check_run_id` → `_run_id_suffix` and `_should_run_check` → `_check_applies_to_resource`; hoists attribute accesses out of inner loop |
| `tests/unit/test_runner.py` | Modified | Updates tests to use `_check_applies_to_resource` and `_ResourceFacts`; adds new test cases for skip-by-name, skip-by-rule-code, and unrelated-skip-checks |
| `tests/unit/test_rule_codes.py` | Modified | Updates `test_skip_checks_with_rule_code` to use the new `_check_applies_to_resource` API and construct `_ResourceFacts` |

### Review Findings
- ✅ No issues with **Python quality** — type hints are present on all new function signatures, imports are clean, `_ResourceFacts` is a well-structured `dataclass(slots=True)`, and the `_pattern_key` helper correctly handles both `str` and `list[str]` inputs.
- ✅ No issues with **check structure** — the refactoring preserves the existing check matching semantics; path filtering (include/exclude) and non-path filtering (materialization + skip_checks) are cleanly separated.
- ✅ No issues with **test coverage** — tests are updated to match the new API. New tests cover materialization mismatch/match, non-model resources, skip-by-name, skip-by-rule-code, and unrelated skip_checks entries (both happy and unhappy paths).
- ✅ No issues with **security** — no deserialization, no secrets, no user-controlled command execution.
- ✅ No issues with **style** — naming follows snake_case, code is consistent with ruff/ruff-format conventions.
- ✅ No issues with **breaking changes** — all changed functions are private (`_`-prefixed). The test imports reference private symbols which is acceptable for unit tests of internal logic.
- ✅ No issues with **documentation** — all new functions have docstrings with Returns sections; inline comments explain the short-circuit optimization rationale.

---

PR #1010, opened 2026-07-30
perf: scan check entry points once per process

pgoslatara:
`importlib.metadata.entry_points()` re-reads the metadata of every installed distribution on each call, a few milliseconds each time, and three call sites needed it during a single run: plugin check discovery, and the two cache-key digests. The installed set cannot change while the process runs, so the scan is now cached for the process lifetime.

Measured on the repo's own config: config assembly **7.09 ms → 0.91 ms**.

Tests that monkeypatch the installed entry-point set need the scan re-run, so `cache_clear()` is added to the existing autouse `_clear_module_caches` fixture.


pgoslatara:
## 🔍 AI PR Review
> **Cache the `entry_points()` scan behind `lru_cache` to avoid redundant metadata reads on every invocation.**

**Type:** `refactor`

### Summary
`importlib.metadata.entry_points()` re-reads the metadata of every installed distribution on each call. Three separate call sites (`_load_entry_point_checks`, `_compute_cache_fingerprint`, `compute_conf_cache_key`) were each calling it independently during a single run. This PR extracts those calls into two `@lru_cache(maxsize=1)` helpers — `_check_entry_points()` and `_check_entry_point_names()` — so the scan happens at most once per process. The test fixture in `conftest.py` is updated to call `.cache_clear()` on both helpers between tests, using a local `_clear()` function to DRY the setup/teardown.

### File Changes
| `src/dbt_bouncer/utils.py` | Modified | Add `_check_entry_points()` and `_check_entry_point_names()` cached helpers; replace three inline `entry_points()` calls with them |
| `tests/conftest.py` | Modified | Clear the new `lru_cache`s between tests; refactor clear logic into a local `_clear()` helper |

### Review Findings
- ✅ No issues with **Python quality** — type hints are correct (`tuple[Any, ...]`, `tuple[str, ...]`), `lru_cache` is already imported, return types are hashable tuples (required for `lru_cache` correctness), docstrings include `Returns` sections.
- ✅ No issues with **check structure** — no check classes were added or modified.
- ✅ No issues with **test coverage** — the `autouse` fixture correctly clears both new caches before and after every test, preventing cross-test pollution. The `_clear()` helper eliminates the previous setup/teardown duplication.
- ✅ No issues with **security** — no deserialization, no user input handling, no secrets.
- ✅ No issues with **style** — snake_case naming, consistent docstring format, clean diff.
- ✅ No issues with **breaking changes** — both new functions are private (`_`-prefixed), and the three replaced call sites are all internal.
- ✅ No issues with **documentation** — both new helpers have complete docstrings. No architecture change requiring AGENTS.md updates.

---

PR #1011, opened 2026-07-30
perf: dispatch on exact type in _wrap_value

pgoslatara:
`_wrap_value` is the hottest function in the codebase — 122,000 calls per run on a 1,000-model project — and performed up to six `isinstance()` calls per value. orjson only ever produces plain `dict`, `list`, `str` and primitives, so an exact-type lookup against a dispatch table is equivalent: the proxy classes are deliberately absent from the table, which is what lets an already-wrapped value fall straight through.

Worth 1–2 ms per run at 1,000 models. Small, but free and confined to one function.

For the record, two larger changes to this file were measured and rejected: writing the wrapped child back into the parent proxy gains nothing (within one run each key is read about once, so the cost is the first wrap, not repeat access), and dropping `ProxyStr` entirely — 58% of all wraps — is only worth 3% and would put `.value` at risk.

Unrelated but unavoidable: ruff 0.16 rejects the `ProxyStr.value` docstring opening with a verb, so `prek` fails on the file without that one-word change.


pgoslatara:
## 🔍 AI PR Review
> **Replaces six `isinstance` calls with a single dict lookup in the hot `_wrap_value` function for a measurable performance improvement.**

**Type:** `refactor`

### Summary
The `_wrap_value` function is called hundreds of thousands of times per run (once for every dict access, list access, and iteration in parsed dbt artifacts). The old implementation used up to six `isinstance` checks (three type checks plus three "not already wrapped" guards). The new version uses a module-level `_WRAPPERS` dispatch dict keyed by exact type (`dict`, `list`, `str`), reducing the per-call cost to a single `dict.get` on `value.__class__`. This is safe because orjson only produces plain `dict`/`list`/`str` (never subclasses), so exact-type dispatch is equivalent to `isinstance` for incoming data, and the proxy subclasses (`DictProxy`, `ListProxy`, `ProxyStr`) are absent from the dispatch table, so already-wrapped values fall through unchanged.

### File Changes
| `src/dbt_bouncer/artifact_parsers/parser.py` | Modified | Replace `isinstance`-based type dispatch in `_wrap_value` with a `dict.get(value.__class__)` lookup against a module-level `_WRAPPERS` table |

### Review Findings
- ✅ No issues with **Python quality** -- type hints are correct (`dict[type, Any]`), no dead code, clean implementation.
- ✅ No issues with **check structure** -- this file is infrastructure, not a check.
- ✅ No issues with **test coverage** -- this is a pure refactor with identical semantics; existing tests cover `_wrap_value` behavior through `DictProxy`/`ListProxy`/`ProxyStr` usage.
- ✅ No issues with **security** -- no deserialization, secrets, or injection vectors.
- ✅ No issues with **style** -- docstrings updated, comment explains the invariant that makes exact-type dispatch safe.
- ✅ No issues with **breaking changes** -- internal function, not part of the public API; behavior is identical.
- ✅ No issues with **documentation** -- docstring accurately describes the optimization rationale.
- ✅ No issues with **cross-file version consistency** -- no version strings involved.

---

PR #1013, opened 2026-07-30
perf: defer the sqlglot import until SQL is actually parsed

pgoslatara:
`sqlglot` costs roughly 65ms to import and was pulled in at module scope by `sql_utils`, and transitively by `checks/manifest/models/code.py`. That cost was paid by every run whose config referenced any of the eight checks defined in `code.py` — even though only three of them (`check_model_does_not_use_cartesian_join`, `check_model_does_not_use_select_star`, `check_model_hard_coded_references`) ever touch the sqlglot AST. A config using only `check_model_max_number_of_lines` was loading a full SQL parser it never invoked.

`sql_utils` now imports sqlglot inside `parse_sql`, matching the deferred-jinja2 pattern already established in that module. `code.py` binds a lazy stand-in for `sqlglot.exp` that replaces the module global with the real module on first attribute access, so the indirection costs nothing after the first hit.

Measured with interleaved cold-interpreter sampling (n=25, alternating A/B in a single loop to cancel machine drift): median import cost of the run path 253ms → 196ms, a saving of ~58ms. This matters more than it sounds — on a realistic project, imports are around two thirds of total wall time, with check execution under 5ms.

Note on `from __future__ import annotations`: it is deliberately *not* added to `code.py`. The `@check` decorator feeds `inspect.signature` annotations into Pydantic `create_model`, so names like `ModelNode` and `Materialization` must stay resolvable at runtime; the two internal helper annotations that reference `exp` are quoted instead.

1214 unit and integration tests pass, and the example config produces identical output (`SUCCESS=720 WARN=4 ERROR=0`).

### DIFF
--- a/src/dbt_bouncer/checks/manifest/models/code.py
+++ b/src/dbt_bouncer/checks/manifest/models/code.py
@@ -1,14 +1,42 @@
 """Checks related to model source code content and structure."""
 
 import re
+from typing import TYPE_CHECKING, Any
 
-from sqlglot import exp
+if TYPE_CHECKING:
+    from sqlglot import exp
 
 from dbt_bouncer.artifact_types import ModelNode
 from dbt_bouncer.check_framework.decorator import check, fail
 from dbt_bouncer.enums import Materialization
 from dbt_bouncer.sql_utils import JINJA_COMMENT_PATTERN, neutralize_jinja, parse_sql
 from dbt_bouncer.utils import compile_pattern, get_clean_model_name
+
+
+class _LazySqlglotExp:
+    """Stand-in for ``sqlglot.exp`` that imports it on first attribute access.
+
+    Only three of this module's checks touch the sqlglot AST, but importing
+    ``sqlglot`` costs ~65ms — paid by every run whose config references *any*
+    check defined here. On first access this swaps itself out for the real
+    module, so the indirection costs nothing after that.
+    """
+
+    def __getattr__(self, name: str) -> Any:
+        """Import ``sqlglot.exp``, rebind the module global, and delegate.
+
+        Returns:
+            Any: The requested attribute of ``sqlglot.exp``.
+
+        """
+        global exp
+        from sqlglot import exp as real_exp
+
+        exp = real_exp
+        return getattr(real_exp, name)
+
+
+exp: Any = _LazySqlglotExp()
 
 # Patterns retained for the best-effort regex fallback used when sqlglot cannot
 # parse a model's SQL (e.g. heavy Jinja control flow).
@@ -56,7 +84,7 @@ def _is_sql_model(model: ModelNode) -> bool:
     return (getattr(model, "language", None) or "sql") == "sql"
 
 
-def _select_uses_star(select: exp.Select) -> bool:
+def _select_uses_star(select: "exp.Select") -> bool:
     """Determine whether a ``SELECT`` projects a star.
 
     Returns:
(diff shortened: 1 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Defer the ~65ms sqlglot import to function-call time so that projects not using AST-based checks avoid the cost entirely.**

**Type:** `refactor`

### Summary
This PR continues the lazy-import performance work started in the prior commits. The top-level `import sqlglot` / `from sqlglot import exp` statements are removed from both `sql_utils.py` and `checks/manifest/models/code.py` and replaced with function-local imports that only execute when SQL is actually parsed or when a check walks the AST. A `TYPE_CHECKING` guard preserves static type resolution so that type checkers still see `exp.Select` / `exp.Expression` as concrete types rather than `Any`. The two files are tightly coupled: `sql_utils.parse_sql` is the single entry point for sqlglot parsing, and `code.py` consumes the returned AST nodes — so deferring the import in both files together is the correct approach.

### File Changes
| `src/dbt_bouncer/sql_utils.py` | Modified | Moved `import sqlglot` and `from sqlglot.errors import SqlglotError` inside `parse_sql()`; moved `from sqlglot import exp` under `TYPE_CHECKING`; added `from __future__ import annotations` to make all annotations lazy (enabling the bare `Environment` return-type annotation). |
| `src/dbt_bouncer/checks/manifest/models/code.py` | Modified | Replaced top-level `from sqlglot import exp` with a `TYPE_CHECKING`-guarded import; added function-local `from sqlglot import exp` in the four functions that reference `exp` types at runtime; quoted the `exp.Select` and `exp.Expression` annotations in function signatures so they are forward references. |

### Review Findings
- ✅ No issues with **Python quality** — `TYPE_CHECKING` guard, `from __future__ import annotations`, and string-quoted annotations in `code.py` (which lacks the future import) are all correctly applied. The `cast()` call in `sql_utils.py:150` already used a string form, so it continues to work.
- ✅ No issues with **check structure** — `@check` decorators, `fail()` calls, and control flow are unchanged.
- ✅ No issues with **test coverage** — this is a pure refactor of import timing with no behavioral change; existing tests cover the affected checks.
- ✅ No issues with **security** — no new deserialization, no secrets, no command injection.
- ✅ No issues with **style** — imports are ordered correctly; snake_case naming preserved.
- ✅ No issues with **breaking changes** — public API (`parse_sql`, check function signatures) is unchanged; the only difference is when the sqlglot module is loaded.
- ✅ No issues with **documentation** — the inline comments clearly explain *why* the import is deferred and why the `TYPE_CHECKING` guard exists.

---

mattcristal:
Super nice ⚡

src/dbt_bouncer/checks/manifest/models/code.py
```diff
@@ -1,15 +1,49 @@
...
+    """
+
+    def __getattr__(self, name: str) -> Any:
+        """Import ``sqlglot.exp``, rebind the module global, and delegate.
+
+        Returns:
+            Any: The requested attribute of ``sqlglot.exp``.
+
+        """
+        global exp
```
mattcristal: I just see it is an anti-pattern to use `global`, but maybe there is no other way 🤔

pgoslatara: Agreed, pushed 91de54b1effd28a12890ee648190416dd8f4892f to address this because it can be avoided.

PR #1014, opened 2026-07-30
perf: drop the twelve redundant resource fields from BaseCheck

pgoslatara:
`BaseCheck` declared a typed field for every resource kind — `catalog_node`, `catalog_source`, `exposure`, `macro`, `model`, `run_result`, `seed`, `semantic_model`, `snapshot`, `source`, `test`, `unit_test` — bringing it to nineteen fields. Pydantic re-collects and copies every inherited field for each subclass, and the codebase generates one subclass per check, so all nineteen were walked ~121 times per run.

Those twelve are redundant. Each check binds exactly one resource and declares that field itself: the `@check` decorator adds it as `Any | None`, and class-based checks must declare it too, because the runner infers `iterate_over` from the subclass's own `__annotations__`. A class-based check that declared none never matched a resource in the first place — it would either match nothing or raise on multiple `iterate_over` values — so nothing that previously worked depends on inheriting them. `set_resource` writes straight into the instance `__dict__`, so no declared slot is needed.

Removing them takes each generated subclass from 19 inherited fields to 7. Measured with interleaved cold-interpreter sampling against `main` (n=15, using a git worktree plus `PYTHONPATH` so both trees stay pristine and no checkout can contaminate a measurement): `validate_conf` on the example config 95.4ms → 84.9ms median, an 11% cut. A synthetic micro-benchmark had suggested closer to 29%; the end-to-end figure is the honest one.

The runtime `artifact_types` import goes with them — it existed only so Pydantic could resolve those field types, and the `_require_*` return annotations are lazy under `from __future__ import annotations`.

`schema.json` regenerates byte-identical, and 1214 unit and integration tests pass.

### DIFF
--- a/src/dbt_bouncer/check_framework/base.py
+++ b/src/dbt_bouncer/check_framework/base.py
@@ -2,24 +2,30 @@
 
 from __future__ import annotations
 
-from typing import Any, ClassVar
+from typing import TYPE_CHECKING, Any, ClassVar
 
 from pydantic import BaseModel, ConfigDict, Field, PrivateAttr
 
-from dbt_bouncer.artifact_types import (  # ruff: ignore[typing-only-first-party-import] - needed at runtime for Pydantic model_rebuild
-    CatalogNodeEntry,
-    ExposureNode,
-    MacroNode,
-    ManifestWrapper,
-    ModelNode,
-    RunResultEntry,
-    SeedNode,
-    SemanticModelNode,
-    SnapshotNode,
-    SourceNode,
-    TestNode,
-    UnitTestNode,
-)
+if TYPE_CHECKING:
+    # Only referenced by the ``_require_*`` return annotations, which are lazy
+    # under ``from __future__ import annotations``. Previously these had to be
+    # imported at runtime so Pydantic could resolve the per-resource field
+    # types; those fields are gone, so the import can go with them.
+    from dbt_bouncer.artifact_types import (
+        CatalogNodeEntry,
+        ExposureNode,
+        MacroNode,
+        ManifestWrapper,
+        ModelNode,
+        RunResultEntry,
+        SeedNode,
+        SemanticModelNode,
+        SnapshotNode,
+        SourceNode,
+        TestNode,
+        UnitTestNode,
+    )
+
 from dbt_bouncer.check_framework.exceptions import DbtBouncerFailedCheckError
 from dbt_bouncer.enums import CheckSeverity, Materialization, RuleCode
 from dbt_bouncer.utils import is_description_populated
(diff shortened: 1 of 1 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Remove twelve redundant per-resource Pydantic fields from `BaseCheck` to cut ~29% of check-class construction cost.**

**Type:** `refactor`

### Summary
Every `BaseCheck` subclass previously inherited twelve `Optional` resource fields (e.g. `model`, `seed`, `source`) even though each check only ever uses one. These fields inflated Pydantic's field-collection overhead during class construction. This PR removes them from the base class — subclasses already declare the one field they need (via the `@check` decorator or manually), and `set_resource()` writes directly into `__dict__` via `object.__setattr__`, so no base-level slot is required. The artifact-type imports are moved behind `TYPE_CHECKING` since they are now only referenced by the `_require_*` return annotations, which are lazy under `from __future__ import annotations`.

### File Changes
| `src/dbt_bouncer/check_framework/base.py` | Modified | Removed 12 per-resource `Field()` declarations from `BaseCheck`; moved artifact-type imports behind `TYPE_CHECKING` |

### Review Findings

- ✅ No issues with **Python quality** — `TYPE_CHECKING` guard is correctly used; `from __future__ import annotations` ensures lazy evaluation of the `_require_*` return annotations; `TYPE_CHECKING` is properly added to the `typing` import.
- ✅ No issues with **check structure** — `set_resource()` uses `object.__setattr__()` which bypasses Pydantic's `extra="forbid"` validation, so runtime assignment of undeclared fields still works. The `_require()` helper uses `getattr(self, field, None)` which correctly reads from `__dict__`. Subclasses (both decorator-generated and class-based) declare their own resource field, so `__annotations__`-based `iterate_over` inference in the runner is unaffected.
- ✅ No issues with **test coverage** — existing tests in `test_check_context.py` exercise `set_resource()` and `model_copy()` on subclasses that declare their own resource field, which is exactly the post-PR pattern. No new code paths were introduced that need new tests.
- ✅ No issues with **security**.
- ✅ No issues with **style** — imports remain alphabetically ordered; the explanatory comment is clear and proportionate to the change.
- ✅ No issues with **breaking changes** — class-based plugin checks that follow the documented pattern (declaring e.g. `model: Any | None = Field(default=None)` on the subclass) are unaffected since those fields were always shadowing the base-class fields anyway. The `@check` decorator adds the field automatically. The only scenario that would break is a plugin that relied on `BaseCheck` having the resource field without declaring it on the subclass — but such a subclass would never have matched a resource in the runner's `__annotations__` scan, so it was already non-functional.
- ✅ No issues with **documentation** — the inline comment thoroughly explains the rationale; no public API surface changed.
---

PR #1019, opened 2026-07-31
test: cover the silent-degradation paths in utils and the conf validator

pgoslatara:
The uncovered code left in `utils.py` and `configuration_file/validator.py` was almost entirely recovery paths: a corrupt check-registry cache, a read-only cache directory, a check module that will not import, an entry point resolving to something unexpected, and a cached conf referencing a check class that is no longer installed. All of them degrade quietly — a wrong turn there costs startup time or silently drops checks rather than raising, so neither CI nor ordinary use would flag a regression.

`test_utils_fallbacks.py` covers the check-registry disk cache (corrupt file rebuild, unwritable directory, warm read), the targeted-import path and its full-scan fallback, a broken internal check module, entry points resolving to a non-module and to a package with an unimportable submodule, and the missing-config-file error.

`test_config_file_validator_cache.py` covers `_load_cached_conf` returning None for every suspect input (absent, unreadable, corrupt JSON, format-version mismatch, unresolved check class), `_write_cached_conf` swallowing both serialisation and OS errors plus its temp-file rename, the `DBT_BOUNCER_DISABLE_CONF_CACHE` escape hatch, and `validate_conf`'s full-scan fallback when no check names can be extracted.

`utils.py` goes 93.1% -> 96.8% and `validator.py` 87.9% -> 93.4%, measured with branch coverage. No source changes.

Two notes for reviewers. `get_check_objects` is `@lru_cache`-decorated and the session-scoped model-rebuild fixture populates it before any test runs, so the test that loads a deliberately broken module set clears the cache either side of itself — `tests/conftest.py::_clear_module_caches` does not currently cover this one. And the two malformed-entry cases assert the resulting `RuntimeError` rather than success: they do reach the full-scan fallback, but the entries are genuinely invalid and validation rejects them, which is the correct behaviour.

### DIFF
--- a/tests/unit/test_config_file_validator_cache.py
+++ b/tests/unit/test_config_file_validator_cache.py
@@ -1,0 +1,247 @@
+"""Tests for the validated-conf disk cache in `dbt_bouncer.configuration_file.validator`.
+
+Every failure mode here is designed to be silent: a bad cache must degrade to a
+full rebuild, never to a wrong config or a crash. That makes these branches
+invisible in normal use and worth pinning down explicitly.
+"""
+
+import logging
+from pathlib import Path
+from types import SimpleNamespace
+
+import orjson
+import pytest
+
+from dbt_bouncer.configuration_file.validator import (
+    _CONF_CACHE_FORMAT_VERSION,
+    _base_field_names,
+    _conf_cache_enabled,
+    _load_cached_conf,
+    _write_cached_conf,
+)
+
+
+def _payload(checks=None, version=_CONF_CACHE_FORMAT_VERSION):
+    return {
+        "v": version,
+        "base": {},
+        "checks": checks if checks is not None else {},
+    }
+
+
+class TestLoadCachedConf:
+    """Tests for `_load_cached_conf`, which must return None on anything suspect."""
+
+    def test_missing_file_returns_none(self, tmp_path):
+        """A cache file that was never written yields a rebuild."""
+        assert _load_cached_conf(tmp_path / "absent.json", set(), None) is None
+
+    def test_unreadable_file_returns_none(self, tmp_path):
+        """A cache path that cannot be read yields a rebuild rather than propagating the OSError."""
+        # A directory exists at the path, so `read_bytes` raises IsADirectoryError.
+        cache_path = tmp_path / "conf.json"
+        cache_path.mkdir()
+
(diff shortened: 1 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Adds unit tests for silent-degradation (error-recovery) paths in the config validator cache and the check-registry utilities.**

**Type:** `test`

### Summary
This PR adds two new test files that cover error-recovery branches in `dbt_bouncer.configuration_file.validator` and `dbt_bouncer.utils`. These branches handle corrupt caches, unwritable directories, unresolvable check classes, and broken entry points — all designed to degrade silently rather than crash. Because they never surface in normal use, they are easy to regress on and worth pinning with explicit tests. The two files are independent but share a common theme: ensuring that every failure path falls back gracefully.

### File Changes
| `tests/unit/test_config_file_validator_cache.py` | New | Tests for `_load_cached_conf`, `_write_cached_conf`, `_conf_cache_enabled`, and `validate_conf` fallback paths |
| `tests/unit/test_utils_fallbacks.py` | New | Tests for check-registry disk cache, targeted-import fallbacks, entry-point loading edge cases, and `load_config_from_yaml` |

### Review Findings
- ✅ No issues with **Python quality** — imports are clean, no type-hint gaps on test helpers, proper use of `pytest.mark.parametrize`, `monkeypatch`, `caplog`, and `tmp_path` fixtures.
- ✅ No issues with **check structure** — these are pure test files; no check classes are defined or modified.
- ✅ No issues with **test coverage** — both happy and unhappy paths are covered; edge cases (corrupt JSON, unwritable paths, version mismatches, unresolvable classes, internal vs. third-party entry points) are all exercised. The `_uncached_get_check_objects` fixture correctly clears the `lru_cache` before and after tests that mutate import behavior.
- ✅ No issues with **security** — no hardcoded secrets, no unsafe deserialization (`orjson` is used for JSON, `yaml.CSafeLoader` in the source is untouched), no command injection vectors.
- ✅ No issues with **style** — snake_case naming throughout, parametrize IDs are descriptive, docstrings on every test method explain intent.
- ✅ No issues with **breaking changes** — no production code is modified; only new test files are added.
- ✅ No issues with **documentation** — test docstrings adequately describe each scenario; no architecture changes requiring AGENTS.md updates.

---

PR #1020, opened 2026-07-31
test: clear the check-registry cache between tests

pgoslatara:
`utils.get_check_objects` is `@lru_cache`-decorated, and the session-scoped `_rebuild_all_check_models` fixture calls it before any test runs. `_clear_module_caches` cleared the two entry-point caches next to it but not this one, so every test started with a pre-warmed registry it had no way to invalidate.

Two consequences. A test that monkeypatches import behaviour and then calls `get_check_objects()` is silently served the stale result — the patch appears to do nothing, which is a confusing failure to diagnose. And a test that deliberately loads a broken registry leaves it cached for whatever runs next.

`tests/unit/test_cache_isolation.py` asserts the fixture's contract for every cache it clears, not just this one, so a cache that stops being cleared fails there rather than surfacing later as an unrelated test reading stale state. Both new assertions fail without the conftest change.

No measurable cost — the fixture only empties the cache, and few tests repopulate it: 1233 unit tests in 4.49s before, 4.48s after.

This makes the local `_uncached_get_check_objects` fixture in #1019 redundant. That fixture is harmless if both merge, and I have left it in place so #1019 stays independent; happy to strip it once this lands.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -62,6 +62,11 @@ def _clear_module_caches():
         # installed set need the scan re-run.
         utils._check_entry_points.cache_clear()
         utils._check_entry_point_names.cache_clear()
+        # The session-scoped model-rebuild fixture populates this before any
+        # test runs, so a test that monkeypatches import behaviour would
+        # otherwise be served a stale registry -- and, worse, could leave a
+        # deliberately broken one behind for whatever runs next.
+        utils.get_check_objects.cache_clear()
 
     _clear()
     yield
--- a/tests/unit/test_cache_isolation.py
+++ b/tests/unit/test_cache_isolation.py
@@ -1,0 +1,45 @@
+"""Tests for the autouse `_clear_module_caches` fixture in `tests/conftest.py`.
+
+Each assertion here holds only because the fixture ran first, whatever executed
+before. A cache that stops being cleared makes exactly one of these fail, rather
+than surfacing as an unrelated test mysteriously reading stale state.
+"""
+
+import pytest
+
+from dbt_bouncer import runner, utils
+from dbt_bouncer.checks.manifest import check_macros
+
+
+@pytest.mark.parametrize(
+    "cache_name",
+    [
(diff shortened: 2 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Clears the `get_check_objects` LRU cache between tests and adds regression tests to guard all cache-clearing logic.**

**Type:** `test`

### Summary
The `_clear_module_caches` autouse fixture in `tests/conftest.py` already cleared four module-level caches between tests, but missed the `get_check_objects` LRU cache. Since the session-scoped `_rebuild_all_check_models` fixture populates that cache once at startup, any test that monkeypatches import behaviour could be served a stale registry — or worse, leave a broken one behind for subsequent tests. This PR adds the missing `cache_clear()` call and introduces a new test file that explicitly verifies all five caches start empty for every test, preventing silent regressions if the fixture drifts out of sync with the caches it should manage.

### File Changes
| `tests/conftest.py` | Modified | Added `utils.get_check_objects.cache_clear()` to the `_clear_module_caches` fixture |
| `tests/unit/test_cache_isolation.py` | New | Parametrised regression tests asserting all LRU and dict caches are empty at test start |

### Review Findings
- ✅ No issues with **Python quality** — type hints present on test signatures, imports are clean and minimal.
- ✅ No issues with **check structure** — no check classes are added or modified.
- ✅ No issues with **test coverage** — tests cover all five caches (3 LRU, 2 dict) with happy paths, plus a dedicated leakage test (`test_populating_a_cache_does_not_leak_out_of_this_test`) that exercises the teardown path. Parametrisation IDs are descriptive.
- ✅ No issues with **security** — no deserialization, secrets, or shell commands.
- ✅ No issues with **style** — snake_case naming, parametrize lists are logically ordered, module docstring present.
- ✅ No issues with **breaking changes** — only test infrastructure is touched; no public API or config schema changes.
- ✅ No issues with **documentation** — the new fixture comment and test docstrings clearly explain the *why*.
---

PR #1018, opened 2026-07-31
test: cover the lazy subcommand loader in `dbt_bouncer.cli`

pgoslatara:
`cli/__init__.py` was the lowest-covered file in the package at 15%. Its `__getattr__` lazy-import loader never fires during a normal run, because `main.py` eagerly imports every subcommand module to trigger `@app.command()` registration. All four resolution branches and the `AttributeError` fallback were therefore untested, and a typo in any of them would have broken third-party `from dbt_bouncer.cli import run` while CI stayed green.

The tests call the loader directly instead of forcing a cold re-import. Purging `dbt_bouncer.cli` from `sys.modules` would construct a second Typer `app`, leaving the subcommands registered against the first one stranded — the test would pass while quietly corrupting the rest of the session.

Now at 100%, statements and branches. No source changes.

One thing surfaced while writing these, worth a separate look: because `init`, `run` and `validate` are also submodule names, importing `dbt_bouncer.main` binds those submodules onto the package and shadows the loader. `from dbt_bouncer.cli import run` yields the *function* on a cold import but the *module* once `dbt_bouncer.main` has been imported. That contradicts the module docstring, which states `__getattr__` serves exactly those callers. `list_checks` is unaffected, having no matching submodule. Not changed here.

PR #1021, opened 2026-07-31
fix: stop re-exporting the cli subcommands that submodules shadow

pgoslatara:
`init`, `run` and `validate` are submodule names as well as subcommand function names. Importing a submodule binds it onto its parent package, and a module-level `__getattr__` is only consulted after normal attribute lookup fails — so the lazy re-export could never win for those three. Because `main.py` eagerly imports all four modules to trigger `@app.command()` registration, `from dbt_bouncer.cli import run` gave you the *function* on a cold import and the *module* once `dbt_bouncer.main` had been touched, which is every CLI invocation.

`__all__` and the module docstring both promised a contract the package could not keep. The docstring went further and claimed `__getattr__` existed to serve "callers that import directly from this package" — the exact case that breaks.

This concedes to Python's semantics rather than fighting them: `init`, `run` and `validate` are dropped from `__all__` and `__getattr__`, and the docstring now explains why and points at `from dbt_bouncer.cli.run import run` and the supported `from dbt_bouncer import run_bouncer`. `list_checks` is kept — no submodule shares its name (the module is `list`), so it resolves reliably. The alternative, rebinding the functions onto the package after the eager import, would mean mutating a package namespace behind the import system's back and would surprise anyone doing `import dbt_bouncer.cli.run`.

The tests added in #1018 are updated in step, asserting the corrected contract: the three shadowed names raise `AttributeError` and resolve to their modules, `list_checks` resolves to the function, and `__all__` claims nothing it cannot deliver.

**Not a breaking change in practice.** `dbt_bouncer.cli` appears in no file under `docs/` or `README.md`; the documented programmatic API is `dbt_bouncer.run_bouncer`. Every internal caller and test already uses the deep path. `src/dbt_bouncer/cli.py` never existed — the `cli/` package was created in `5f331f29` and first shipped in v3.1, so this was never preserving a pre-v3 import path. Under any realistic import order you already got the module, so nothing observable changes; only a cold import that touches `dbt_bouncer.cli` before `dbt_bouncer.main` differs, and there a function becomes a module and the call site raises immediately rather than failing silently. Suitable for a patch release.

Verified beyond the suite: `dbt-bouncer --help` still lists all four subcommands, and a full run against `dbt-bouncer-example.yml` still reports `SUCCESS=720 WARN=4 ERROR=0`.

### DIFF
--- a/src/dbt_bouncer/cli/__init__.py
+++ b/src/dbt_bouncer/cli/__init__.py
@@ -3,9 +3,26 @@
 The ``app`` Typer instance lives here so that each subcommand module can
 register itself with ``@app.command()`` at import time.
 
-``main.py`` eagerly imports each subcommand module to trigger
-registration, so ``__getattr__`` only fires for callers that import
-directly from this package (e.g. ``from dbt_bouncer.cli import init``).
+``list_checks`` is re-exported lazily via ``__getattr__``, so importing this
+package does not pull in the subcommand modules.
+
+The other three subcommand functions are deliberately *not* re-exported.
+``init``, ``run`` and ``validate`` are also submodule names, and importing a
+submodule binds it onto its parent package — which shadows a module-level
+``__getattr__``, since that is only consulted after normal attribute lookup
+fails. ``main.py`` eagerly imports all four modules to trigger registration,
+so in any real process ``dbt_bouncer.cli.run`` is the *module*. Re-exporting
+the functions made the result depend on import order: the function on a cold
+import, the module once ``dbt_bouncer.main`` had been touched.
+``list_checks`` is unaffected only because no submodule shares its name.
+
+Reach the other subcommands through their own modules instead::
+
+    from dbt_bouncer.cli.run import run
+
+For programmatic use prefer the supported entry point, ``run_bouncer``::
+
+    from dbt_bouncer import run_bouncer
 """
 
 import typer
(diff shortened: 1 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Fixes import-order-dependent behavior by stopping `dbt_bouncer.cli` from re-exporting subcommand functions whose names collide with submodule names.**

**Type:** `bugfix`

### Summary
`dbt_bouncer.cli` has submodules named `init`, `run`, and `validate` — the same names as the subcommand functions they contain. When `main.py` eagerly imports those submodules for Typer registration, Python binds the *module* onto the parent package, permanently shadowing the `__getattr__`-based lazy re-exports. The result was import-order-dependent: callers got the function on a cold import but the module after `main.py` had run. This PR removes the three unreliable re-exports (keeping only `list_checks`, whose module is named `list` and is therefore unaffected) and updates the tests to verify the new, deterministic behavior.

### File Changes
| `src/dbt_bouncer/cli/__init__.py` | Modified | Removed `init`, `run`, `validate` from `__getattr__` and `__all__`; expanded module docstring explaining the shadowing issue |
| `tests/unit/test_cli_lazy_imports.py` | Modified | Replaced parametrised happy-path test with individual tests: `list_checks` resolves, shadowed names raise `AttributeError` from `__getattr__`, and after `main` import they resolve to their submodules |

### Review Findings
- ✅ No issues with **Python quality** — type hints present, clean imports, no dead code.
- ✅ No issues with **test coverage** — tests cover the happy path (`list_checks` resolves to callable), the shadowing behavior (three names raise `AttributeError` from `__getattr__`), the submodule binding (`getattr` yields `ModuleType` after `main` import), and the `__all__` contract. Edge cases for unsupported names and `app` attribute are preserved.
- ✅ No issues with **style** — `__all__` is alphabetically ordered, naming follows conventions.
- ✅ No issues with **breaking changes** — grep confirms no code in the repo uses `from dbt_bouncer.cli import init/run/validate`; all existing callers import from the submodules directly. The public entry point `run_bouncer` is unaffected.
- ✅ No issues with **documentation** — module docstring thoroughly explains the problem and recommends the correct import paths.
- ✅ No issues with **cross-file version consistency** — not applicable to this PR.
---

PR #1027, opened 2026-08-04
tests: expand and harmonize unit tests for model checks

mattcristal:
# What was done

## Why

This PR would solve [this open issue](https://github.com/godatadriven/dbt-bouncer/issues/961)

## Summary

Screenshot 2026-08-04 at 08 17 17

## Harmonization

`models/` had never received #940's treatment. All 11 files now use the `check_fn` (check_passes/check_fails) column convention, with pytest.raises/caplog/message-assertion methods left standalone per #940's precedent. Worst offenders collapsed hardest: test_lineage.py went 17 classes/36 methods → 11 classes (one per check, with the …InvalidParam classes folded in), and repetitive model dicts moved behind small builders.

PR #1028, opened 2026-08-04
fix: correct constant-ON and NATURAL JOIN detection in cartesian join check

pgoslatara:
`check_model_does_not_use_cartesian_join` had two independent false-positive/false-negative sources in its AST path, both surfaced while reviewing #1027 (which pinned the buggy behaviour as expected, with comments noting it looked wrong).

**Constant-`ON` detection was asymmetric.** `join.args["on"]` holds the join condition itself, but the code descended one level further into `on_clause.this`. That is correct only for `ON TRUE`/`ON FALSE`, where sqlglot produces an `exp.Boolean` whose `.this` is a Python `bool` — the case the old comment defended. For every other shape it inspected only the predicate's left operand, so:

| SQL | Before | After |
|---|---|---|
| `ON 1 = b.id` | flagged as Cartesian | passes |
| `ON b.id = 1` | passes | passes |
| `ON 1=1` | fails, message reads `` `ON 1` `` | fails, message reads `` `ON 1 = 1` `` |
| `ON 1` | **not flagged** | flagged |

The `isinstance(cond, exp.EQ)` sub-check was unreachable for the same reason — `cond` was already the left operand of that `EQ`. The condition is now evaluated as a whole in `_constant_join_condition`, which treats a comparison as constant only when both operands are literals.

**`NATURAL JOIN` was a false positive.** It carries neither an `ON` nor a `USING` clause, so it hit the missing-clause branch and was reported as "a `JOIN` without an `ON` or `USING` clause". It joins on the columns the two relations share, so it is now exempt via `method == "NATURAL"`. Implicit comma joins (`FROM a, b`) also carry no clause but *are* genuine Cartesian products, and still fail.

`ON NULL` remains unflagged — it matches no rows, so it is degenerate rather than Cartesian.

The three parametrized cases in `test_code.py` that documented the old behaviour are inverted, and cases are added for the bare-literal `ON 1` false negative, equal string literals, the implicit comma join, and the `ON TRUE` failure message. `schema.json` is regenerated for the amended docstring.


pgoslatara:
## 🔍 AI PR Review
> **Fixes two false positives (constant-ON asymmetry and NATURAL JOIN) in the Cartesian join check, extracts the condition logic into a testable helper, and adds comprehensive test coverage.**

**Type:** `bugfix`

### Summary
The constant-condition detection in `check_model_does_not_use_cartesian_join` previously inspected only the left operand of an `ON` clause (`on_clause.this`), causing `ON 1 = b.id` to be flagged while the equivalent `ON b.id = 1` passed. It also flagged `NATURAL JOIN` as a missing-clause join, even though `NATURAL JOIN` implicitly joins on shared columns. This PR extracts the condition logic into a standalone `_constant_join_condition()` helper that checks both operands, adds `NATURAL JOIN` detection, and updates the schema description and tests to match.

The three files are tightly coupled: `code.py` contains the fix, `test_code.py` validates it (moving previously-failing cases to the passing suite and adding new edge cases), and `schema.json` is the auto-generated schema reflecting the updated docstring.

### File Changes
| `src/dbt_bouncer/checks/manifest/models/code.py` | Modified | Extracted `_constant_join_condition()` helper; added `NATURAL JOIN` detection; replaced inline constant-condition logic with helper call |
| `tests/unit/checks/manifest/models/test_code.py` | Modified | Added passing cases for `NATURAL JOIN`, literal-on-either-side, columns-on-both-sides; added failing cases for bare literal, string literals, implicit comma join; renamed and updated symmetry test; fixed expected error message patterns |
| `schema.json` | Modified | Updated `CheckModelDoesNotUseCartesianJoin` description to document symmetric detection, `NATURAL JOIN` handling, and clarified constant-condition semantics |

### Review Findings
- ✅ No issues with **Python quality** — the new `_constant_join_condition` helper has proper type hints, a clear docstring with Args/Returns, and the lazy `from sqlglot import exp` import follows the existing pattern in the file.
- ✅ No issues with **check structure** — `@check` decorator, `fail()` usage, and `iterate_over_value` are unchanged and correct.
- ✅ No issues with **test coverage** — all three bug-fix scenarios are covered: `NATURAL JOIN` (passing), literal on either side of a predicate (both passing), implicit comma join (failing), bare literal and string-literal constants (failing). The error message assertions are updated to match the new rendering (`ON 1 = 1` instead of `ON 1`). The symmetry test is renamed to accurately reflect its purpose.
- ✅ No issues with **security** — no deserialization, no user-controlled shell commands, no secrets.
- ✅ No issues with **style** — consistent with existing code conventions, snake_case naming, docstring formatting.
- ✅ No issues with **breaking changes** — the check name, code (`MO052`), and config parameters are unchanged. The only behavioral changes are removing false positives (previously-flagged valid SQL now passes), which is non-breaking.
- ✅ No issues with **documentation** — docstring updated in the check function and accurately reflected in `schema.json`.

---

PR #1030, opened 2026-08-04
fix: report undocumented models as undocumented, not misplaced

pgoslatara:
`check_model_documented_in_same_directory` guarded its "is not documented" failure with a condition that can never be false:

```python
if not (hasattr(model, "patch_path")
        and clean_path_str(model.patch_path or "") is not None):
```

`clean_path_str` is annotated `-> str` and returns `""` rather than `None` (`utils.py:36`), and `model.patch_path or ""` already coerces `None` to `""`, so `is not None` always holds. `patch_path` is a declared field on the wrapped resource, so `hasattr` always holds too. The branch was unreachable — it was the only uncovered line left in the module.

An undocumented model instead fell through to the directory comparison, where `Path("").parent.parts` is `()`, and was reported as:

> `model_1` is documented in a different directory to the `.sql` file: `` vs `models/staging`.

which points a user at a directory mismatch that does not exist. The guard now tests the cleaned patch_path for emptiness, which is what the original condition was reaching for.

Undocumented models still fail, exactly as before — only the message changes — so this is not a behaviour change for passing or failing projects, just a correct diagnosis. The three parametrized cases in `test_description.py` that pinned the misleading message (absent, `None`, and empty-string `patch_path`) now assert the intended one.


pgoslatara:
## 🔍 AI PR Review
> **Fixes `check_model_documented_in_same_directory` to correctly report undocumented models as "not documented" instead of misleadingly reporting them as "documented in a different directory".**

**Type:** `bugfix`

### Summary
The old code used `hasattr(model, "patch_path")` and `clean_path_str(...) is not None` to detect undocumented models, but `clean_path_str` always returns a `str` (never `None`), so the "is not documented" branch was unreachable. Undocumented models fell through to the directory comparison and were reported with a confusing "different directory" message against an empty path. This PR consolidates the logic into a single `getattr` + falsy check, making the undocumented branch reachable and producing the correct error message. The test and schema changes align with the fix.

### File Changes
| `src/dbt_bouncer/checks/manifest/models/description.py` | Modified | Rewrote the undocumented-model guard to use `getattr` with a falsy check instead of the broken `hasattr`/`is not None` combo; added docstring note |
| `tests/unit/checks/manifest/models/test_description.py` | Modified | Updated test name, comment, and expected error message to match the corrected behavior |
| `schema.json` | Modified | Regenerated schema to include the new `!!! info` docstring note about undocumented models |

### Review Findings
- ✅ No issues with **Python quality** — the `getattr(..., None) or ""` pattern is idiomatic; `fail()` is typed `NoReturn` so execution correctly halts before `patch_path_str` is reused on line 222.
- ✅ No issues with **check structure** — `@check` decorator and `fail()` used correctly.
- ✅ No issues with **test coverage** — three parametrised unhappy-path cases (`patch_path` absent, `None`, empty string) all assert the corrected error message; happy-path tests for same-directory documentation are already present above.
- ✅ No issues with **security** — no deserialization, no secrets, no injection vectors.
- ✅ No issues with **style** — changes are minimal and consistent with surrounding code conventions.
- ✅ No issues with **breaking changes** — the check name, code (`MO023`), and config schema are unchanged; only the error message text changed, which is not a stable API.
- ✅ No issues with **documentation** — docstring updated with the `!!! info` admonition; schema regenerated to match.
---

PR #1035, opened 2026-08-04
docs: disambiguate the include parameter of check_model_directories

pgoslatara:
Fixes the contradictory documentation of `check_model_directories`'s `include` parameter, which was documented twice with different types and meanings (required `str` "directory to check" in Parameters vs the boilerplate optional `str | list[str]` path filter in Other Parameters).

The check's keyword-only `include: str` overrides the inherited `BaseCheck.include` field, so one value plays both roles: it is the runner-level path filter AND the directory scope the check validates against. The docstring now describes that dual role explicitly and the stale duplicate entry is gone. `schema.json` is regenerated (description field only); no behaviour change.

Part of the v4 preparation series (5/10).


pgoslatara:
## 🔍 AI PR Review
> **Clarifies the `include` parameter documentation for `check_model_directories` to distinguish it from the standard path-filter `include` used by other checks.**

**Type:** `docs`

### Summary
The `check_model_directories` check uses `include` as a required, single-string regex that controls which directory prefix is matched — fundamentally different from the optional list-of-patterns `include` that other checks use for path filtering. The old docstring described it with the same one-liner as every other check, which was confusing. This PR rewrites the `include` parameter description to explain its unique semantics and removes the duplicate `include` entry from the `Other Parameters` section (which incorrectly described it as the standard optional list form). The `schema.json` file is regenerated to reflect the updated docstring.

### File Changes
| `src/dbt_bouncer/checks/manifest/models/directories.py` | Modified | Rewrote `include` param docstring to clarify its check-specific semantics; removed the misleading duplicate `include` line from `Other Parameters` |
| `schema.json` | Modified | Regenerated to reflect the updated docstring for `CheckModelDirectories` |

### Review Findings
- ✅ No issues with **Python quality** — no code logic changed, only docstring text.
- ✅ No issues with **Check structure** — decorator, `fail()` usage, and function signature are unchanged.
- ✅ No issues with **Test coverage** — documentation-only change; no new behaviour to test.
- ✅ No issues with **Security** — no runtime changes.
- ✅ No issues with **Style** — docstring formatting is consistent with surrounding checks.
- ✅ No issues with **Breaking changes** — the `include` parameter type and runtime behaviour are unchanged; only the docstring and generated schema description were updated.
- ✅ No issues with **Documentation** — the expanded description accurately reflects the actual behaviour of the `include` parameter in `check_model_directories`.
- ✅ No issues with **Cross-file version consistency** — not applicable (no version bumps).
---

PR #1042, opened 2026-08-04
ci: authenticate mise's GitHub API calls in the dev-container job

pgoslatara:
The `dev-container` job fails intermittently on GitHub API rate limits rather than on anything in the change under test. `mise` resolves `ubi:` tools (currently `ubi:bencherdev/bencher`) through the GitHub API, and with no token in the container it makes unauthenticated requests, which are limited to 60/hour **per IP** and shared across every job on that runner address.

A recent run failed exactly this way while the same job had passed on the previous commit twenty minutes earlier, with the only change in between being a test file:

```
mise WARN  No GitHub token was found, so mise is making unauthenticated requests to GitHub which have a much lower rate limit.
github rate limit: 0/60 (core), resets at 1785851282
mise ERROR Failed to install ubi:bencherdev/bencher@0.6.6: HTTP status client error (403 Forbidden)
```

This passes the workflow's existing `GITHUB_TOKEN` into the container via the `devcontainers/ci` action's `env` input, which mise picks up automatically — the log itself points at this fix. Authenticated requests get a much higher, per-account limit, so the job's outcome depends on the code again rather than on how busy the shared runner IP has been.

The token needs no scopes for these read-only API calls, and the job now declares `permissions: contents: read` explicitly so it is least-privilege rather than inheriting the workflow default. On fork pull requests the token is read-only, so this does not widen what a fork PR can do.

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -187,6 +187,8 @@ jobs:
 
     dev-container:
       needs: [prek]
+      permissions:
+        contents: read
       runs-on: ubuntu-24.04
       steps:
         - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
@@ -196,6 +198,12 @@ jobs:
         - name: Build container and run tests
           uses: devcontainers/ci@513af61f4de4f75d37e4438f184ba4358f0fc1ca # v0.3
           with:
+            # mise resolves `ubi:` tools via the GitHub API. Unauthenticated
+            # runners share an IP-based limit of 60 requests/hour, so without a
+            # token this job fails intermittently on rate limits rather than on
+            # anything in the change under test.
+            env: |
+              GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}
             runCmd: |
               set -eou
               uv run dbt-bouncer --config-file dbt-bouncer-example.yml


pgoslatara:
## 🔍 AI PR Review
> **Authenticates mise's GitHub API calls in the dev-container CI job to prevent intermittent rate-limit failures.**

**Type:** `chore`

### Summary
The `dev-container` job in the CI pipeline uses mise to resolve `ubi:` tools via the GitHub API. Without authentication, runners share an IP-based rate limit of 60 requests/hour, causing intermittent failures unrelated to the code under test. This PR passes `GITHUB_TOKEN` into the devcontainer environment and adds the minimum `contents: read` permission the job needs.

### File Changes
| `.github/workflows/ci_pipeline.yml` | Modified | Adds `permissions: contents: read` to the `dev-container` job and passes `GITHUB_TOKEN` as an env var to the devcontainer via the `devcontainers/ci` action's `env` input |

### Review Findings
- ✅ No issues with **security** — `secrets.GITHUB_TOKEN` is the automatic, scoped GitHub token (not a PAT or hardcoded secret), and the job's permissions are narrowed to `contents: read`, which is the minimum needed. The checkout step already sets `persist-credentials: false`.
- ✅ No issues with **style** — the inline comment clearly explains *why* the token is needed.
- ✅ No issues with **breaking changes** — this is an additive CI configuration change with no effect on the package or its public API.
- ✅ No issues with **cross-file consistency** — other workflow files already use `secrets.GITHUB_TOKEN` in similar patterns; no stale references found.
- ✅ Remaining categories (Python quality, check structure, tests, documentation) are not applicable — this PR only modifies a GitHub Actions workflow file.
---

PR #1043, opened 2026-08-04
Resolving dbt-bouncer warnings

pgoslatara:
Resolves #

### Checklist

- [ ] I have run this code in development, and it appears to resolve the stated issue.
- [ ] This PR includes tests, or tests are not required or relevant for this PR.

PR #1033, opened 2026-08-04
refactor!: trim back-compat re-exports from the public API

pgoslatara:
Trims the back-compat re-export layers from the public API for the v4 major release.

- Removes the `run_bouncer` self-alias re-export from `main.py` — the canonical paths are `dbt_bouncer.run_bouncer` (lazy, unchanged) and `dbt_bouncer.cli.run.utils.run_bouncer`.
- Removes `cli/__init__.py`'s lazy `__getattr__` and its `list_checks` re-export (`from dbt_bouncer.cli.list import list_checks` is now the only path), shrinking the 17-line shadowing docstring to three lines; `__all__ = ["app"]`.
- `dbt_bouncer/__init__.py`'s `__all__` now derives from the single `_ENUM_NAMES` source instead of a hand-synced duplicate list — same nine public names, and the PEP 562 lazy loader stays (it is a startup-performance feature).
- The RUF067 per-file-ignore stays: without it ruff flags the six `__init__.py` files that legitimately hold runtime code (Typer command registration, the lazy loader); its comment now says so.

Tests: 14 obsolete lazy-import tests removed (they covered the deleted `__getattr__` mechanism), 5 added covering the new public surface. Full suite passes at 1633 unit and 20 integration tests.

Part of the v4 preparation series.

PR #1032, opened 2026-08-04
refactor!: drop the class-based check framework

pgoslatara:
Removes the legacy class-based check framework for the v4 major release — the `@check` decorator is now the only way to define a check. Zero built-in checks used the class-based path, and the docs have described it as a legacy escape hatch since v3.

Stacked on #1031 (the `check_patterns` shim re-exported the pattern ABCs removed here); the base branch will be updated to `main` once #1031 merges.

- Deletes `check_framework/patterns.py` (six unused `Base*Check` ABCs whose logic had already drifted from the live checks) and its tests.
- Deletes the dead `_require`/`_require_*` accessor layer from `BaseCheck` (~140 lines, no callers). `set_resource` stays — the executor binds resources through it.
- Simplifies runner dispatch to the `iterate_over` ClassVar set by `@check`, removing the `__annotations__`-intersection fallback and the now-unreachable "multiple iterate_over values" error. `_CLASS_ITERATE_CACHE` stays for the `--dry-run` table.
- Docs: removes the class-based sections from CONTRIBUTING.md and AGENTS.md; the v3 migration guide keeps its historical narrative and notes the removal in v4.

Custom checks that subclass `BaseCheck` with `execute()` are no longer discovered; migrate them to `@check` per the contributing guide. `schema.json` is unchanged.

Part of the v4 preparation series (2/10).

PR #1044, opened 2026-08-04
fix: detect dbt built-in macro overrides from their own dispatch edge

pgoslatara:
`check_macro_is_used` fails on project macros that override a dbt built-in — e.g. `generate_schema_name` — when the manifest is generated by dbt-core 2.0.

Such macros are invoked by dbt itself, never by another resource, so they appear in no `depends_on.macros` anywhere in the manifest. The check compensated by scanning the `dbt` package for the dispatch edge `macro.dbt.<name>` → `macro.dbt.default__<name>` and treating any macro sharing one of those names as used. dbt 2.0 no longer populates `depends_on.macros` for macros in the built-in `dbt` and adapter packages, so that lookup yields nothing and every override is reported as unused.

Macros with a non-empty `depends_on.macros`, per package, from `dbt compile` on this repo's `dbt_project/`:

| package | dbt 1.12.0 | dbt 2.0.0a5 |
|---|---|---|
| `dbt` | 287 / 423 | **0 / 450** |
| `dbt_duckdb` | 37 / 63 | **0 / 71** |
| `dbt_date` | 150 / 173 | 150 / 173 |
| `dbt_expectations` | 90 / 135 | 90 / 135 |
| `dbt_utils` | 83 / 116 | 83 / 116 |
| `fivetran_utils` | 32 / 93 | 32 / 93 |
| `spark_utils` | 13 / 20 | 13 / 20 |
| `tiktok_ads` | 13 / 15 | 13 / 15 |

The loss is confined to the built-in `dbt` and adapter packages; every installed hub package is identical across both versions. That is consistent with dbt 2.0 resolving global and adapter macros internally rather than parsing them into the manifest graph.

A macro is now also treated as used when it depends on a dispatched implementation of its own name — `default__<name>` or `<adapter>__<name>`. Those edges are recorded on the *overriding* macro under both 1.x and 2.0 (`macro.dbt_bouncer_test_project.generate_schema_name` → `["macro.dbt.default__generate_schema_name"]`), so detection no longer depends on the built-in packages' own graph.

The existing `dbt`-package heuristic is kept rather than replaced, so nothing regresses under dbt 1.x for an override that never calls the dispatched implementation. That case has no signal at all in a 2.0 manifest and remains unsolved here.

## Verification

Against a real manifest built with `dbt-core==2.0.0a5` + `dbt-duckdb~=1.10.0` (`dbt compile --write-catalog`), `check_macro_is_used` alone:

| | before | after |
|---|---|---|
| dbt 2.0.0a5 | `ERROR=1` on `generate_schema_name` | `SUCCESS=3 ERROR=0` |
| dbt 1.10 / 1.11 / 1.12 fixtures | `SUCCESS=3 ERROR=0` | `SUCCESS=3 ERROR=0` |

Plus 1555 unit and 19 integration tests. Rebased onto `42dbafb4`; re-verified after #1032 (class-based framework dropped), #1033 (public API trimmed) and #1043 (fixtures regenerated, dbt 1.9 fixtures deleted).

## Caveat: `dbt parse` artifacts on 2.0 are not covered

On dbt 2.0, `dbt parse` emits **no** macro-to-macro `depends_on` edges at all — 0 / 1076 across every package, including hub packages and the project's own macros. `dbt compile` populates them (the table above). dbt 1.12 `dbt parse` populates them fully, so this is a 2.0 behaviour change, not a parse-vs-compile distinction that already existed.

Node-to-macro edges survive `dbt parse` on 2.0 (81 / 109 nodes), so a macro called directly by a model is still detected. A macro reachable only from another macro is not, and this fix cannot help — the dispatch edge it keys on is absent too. The `build-artifacts-1XX` tasks in `mise.toml` use `dbt parse` followed by `dbt docs generate`; on 2.0, `dbt docs generate` is rejected in favour of `dbt compile --write-catalog`, so a future `build-artifacts-20` task needs `dbt compile`, not `dbt parse`, or the fixtures will carry an empty macro graph.

Draft because dbt-core 2.0 is still alpha and there is no `dbt_20` fixture committed, so the 2.0 shape is covered by unit tests transcribing a real 2.0 manifest rather than by an artifact in the repo.

### DIFF
--- a/src/dbt_bouncer/checks/manifest/check_macros.py
+++ b/src/dbt_bouncer/checks/manifest/check_macros.py
@@ -497,6 +497,36 @@ def check_macro_property_file_location(macro):
 _USED_MACROS_CACHE: dict[int, set[str]] = {}
 
 
+def _dispatches_to_own_name(macro: Any) -> bool:
+    """Whether a macro calls a dispatched implementation of its own name.
+
+    dbt's dispatch convention names implementations `<prefix>__<name>`, e.g.
+    `default__generate_schema_name` or `duckdb__apply_grants`. A macro that
+    depends on such an implementation of its own name is overriding a dbt
+    built-in, and is therefore invoked by dbt itself rather than by another
+    resource in the manifest.
+
+    Args:
+        macro: The macro object to inspect.
+
+    Returns:
+        True if the macro overrides a dispatched built-in.
+
+    """
+    depends_on = getattr(macro, "depends_on", None)
+    if depends_on is None or not hasattr(depends_on, "macros"):
+        return False
+
+    suffix = f"__{macro.name}"
+    for unique_id in depends_on.macros:
+        # `unique_id` is `macro.<package_name>.<macro_name>`.
+        dispatched_name = unique_id.rsplit(".", 1)[-1]
+        # `!= suffix` rejects an empty dispatch prefix.
+        if dispatched_name.endswith(suffix) and dispatched_name != suffix:
+            return True
+    return False
+
+
 def _get_used_macros(manifest_obj: Any) -> set[str]:
     obj_id = id(manifest_obj)
     if obj_id not in _USED_MACROS_CACHE:
@@ -516,7 +546,10 @@ def _get_used_macros(manifest_obj: Any) -> set[str]:
                 if hasattr(item, "depends_on") and hasattr(item.depends_on, "macros"):
                     used_macros.update(item.depends_on.macros)
 
-        # Add macros that override default dbt macros (identifiable via the depends_on key-value pair)
+        # Add macros that override default dbt macros. dbt < 2.0 records the
+        # dispatch edge on the built-in itself (`macro.dbt.<name>` depends on
+        # `macro.dbt.default__<name>`), so the overridable names can be read
+        # straight off the `dbt` package.
         overridable_dbt_macros = set()
         macros_collection = getattr(manifest_data, "macros", {})
         for item in macros_collection.values():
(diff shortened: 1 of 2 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Fixes false-positive "macro is unused" warnings for dbt built-in macro overrides when running against dbt 2.0 manifests.**

**Type:** `bugfix`

### Summary
dbt 2.0 no longer populates `depends_on.macros` for macros in the built-in `dbt` and adapter packages. This broke the existing detection logic that identified overridable dbt macros by inspecting `depends_on` on `macro.dbt.<name>` entries, causing user-defined overrides (e.g. `generate_schema_name`, `apply_grants`) to be incorrectly flagged as unused. This PR adds a new helper `_dispatches_to_own_name()` that detects overrides from the *overriding* macro's own dispatch edge instead, making detection work regardless of dbt version.

The two files are tightly coupled: the implementation adds the new detection path in `check_macros.py`, and the tests in `test_macros.py` add three new parametrised cases covering the dbt 2.0 scenario, adapter-dispatched builtins, and a negative case ensuring non-override dispatch calls still fail.

### File Changes
| `src/dbt_bouncer/checks/manifest/check_macros.py` | Modified | Adds `_dispatches_to_own_name()` helper and integrates it as a fallback in `_get_used_macros()` for dbt 2.0 compatibility |
| `tests/unit/checks/manifest/test_macros.py` | Modified | Adds 3 parametrised test cases: dbt 2.0 override detection, adapter-dispatched override, and negative case for unrelated dispatch calls |

### Review Findings
- ✅ No issues with **Python quality** — function has proper type hint (`macro: Any`), docstring with Args/Returns, clean logic. The `Any` type is consistent with how other helpers in this file handle macro objects.
- ✅ No issues with **check structure** — the change is in a private helper consumed by the existing `MA005` check; no decorator or `BaseCheck` changes needed.
- ✅ No issues with **test coverage** — three new parametrised cases cover the key scenarios: happy path (dbt 2.0 override without dbt-package dispatch edge), happy path (adapter-dispatched override like `duckdb__apply_grants`), and unhappy path (calling a dispatched macro that doesn't match the caller's own name). Edge cases around empty prefixes are handled by the `!= suffix` guard in the code.
- ✅ No issues with **security** — no deserialization, no external input, no secrets.
- ✅ No issues with **style** — code follows existing conventions, snake_case naming, comments explain the "why" clearly.
- ✅ No issues with **breaking changes** — purely additive; existing behavior for dbt < 2.0 is unchanged, new fallback only activates when the existing detection yields an empty set.
- ✅ No issues with **documentation** — the new helper has a complete docstring; inline comments explain the dbt version differences well. No architecture changes requiring AGENTS.md updates.
---

PR #1046, opened 2026-08-07
fix: run build-artifacts sub-tasks in series to avoid DuckDB lock contention

pgoslatara:
`mise run build-artifacts` intermittently failed with a DuckDB lock error. The aggregate task ran its version-specific sub-tasks (`build-artifacts-110`, `-111`, `-112`, `-local`) in parallel through mise `depends`. Each sub-task isolates its compiled artifacts with `--target-path`, but they all share one database file (`path: ./dbt.duckdb` in `profiles.yml`). The `dbt docs generate` step opens that file read-write and DuckDB takes an exclusive lock, so the parallel builds fought over it.

This changes the aggregate to run the sub-tasks in series via a `run` list. The build is slower but no longer contends on the shared database. The comment now records why series is the default and what would be needed to parallelise safely (a separate DuckDB file per version).

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -44,18 +44,20 @@ run = [
   "uv run dbt-bouncer --config-file ./dbt-bouncer-example.yml",
 ]
 
-# Each version-specific task uses --target-path to write to an isolated
-# directory, so the version builds (110, 111, 112) run in parallel via the
-# `build-artifacts` aggregate. Parallel execution may cause conflicts if dbt
-# acquires project-level locks in ./dbt_project. If that happens, run with
-# `mise run --jobs 1 build-artifacts`.
+# Each version-specific task uses --target-path to write compiled artifacts to
+# an isolated directory. The DuckDB database is NOT isolated: every task shares
+# the same `path: ./dbt.duckdb` from profiles.yml. The `dbt docs generate` step
+# opens that file read-write, and DuckDB takes an exclusive lock. Running the
+# version builds in parallel makes them fight for that lock, so this aggregate
+# runs the sub-tasks in series. This is slower but reliable. To parallelise,
+# give each version its own DuckDB file first.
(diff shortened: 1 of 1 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Fixes DuckDB lock contention by running `build-artifacts` sub-tasks in series instead of in parallel.**

**Type:** `bugfix`

### Summary
The `build-artifacts` mise task previously used `depends` to declare its sub-tasks (`build-artifacts-110`, `build-artifacts-111`, `build-artifacts-112`, `build-artifacts-local`), which mise executes in parallel by default. Because all sub-tasks share the same DuckDB database file (`dbt.duckdb` from `profiles.yml`), the `dbt docs generate` step in each sub-task fights for an exclusive DuckDB lock, causing intermittent failures. This PR switches from `depends` (parallel) to a sequential `run` array that invokes each sub-task via `mise run`, ensuring they execute one at a time. The comment block is also rewritten to clearly explain the root cause and how to re-enable parallelism in the future.

### File Changes
| `mise.toml` | Modified | Replace `depends` (parallel execution) with sequential `run` commands for `build-artifacts` task; update explanatory comment |

### Review Findings
- ✅ No issues with **Python quality** (no Python code changed)
- ✅ No issues with **check structure** (no check code changed)
- ✅ No issues with **test coverage** (infrastructure-only change to task runner config)
- ✅ No issues with **security** (no secrets, no deserialization, no command injection)
- ✅ No issues with **style** (comment is clear, sub-tasks listed in consistent order)
- ✅ No issues with **breaking changes** (same sub-tasks run in same order, just sequentially)
- ✅ No issues with **documentation** (the inline comment thoroughly explains the why and how-to-parallelize escape hatch)
- ✅ No issues with **cross-file version consistency** (no version bumps in this PR)
---

PR #1056, opened 2026-08-12
ci: bump jdx/mise-action to v4.2.4 and let Dependabot scan composite actions

pgoslatara:
The pinned `jdx/mise-action@v2.4.4` bundles a Node.js 20 runtime, which GitHub deprecated. Every CI job routes through the `setup_python_env` composite action, so each job emitted an identical Node.js 20 deprecation annotation — 14 warnings per run. `v4.2.4` targets Node.js 24 and clears them. The `install` input this action passes is unchanged in `v4`, so the major-version bump does not affect the config. CI is the real test of the upgrade.

Dependabot never raised this bump because its `github-actions` updater with `directory: /` scans `.github/workflows` and a root `action.yml`, but not composite actions nested under `.github/actions`. `jdx/mise-action` is the only action referenced from a composite file, so it is the only one that went stale while every workflow-level action stayed current.

This PR also fixes that gap. It switches the `github-actions` config to `directories` and adds a `/.github/actions/*` glob, so nested composite-action pins now get update PRs. Both composite actions (`setup_python_env` and `resolve-pr-from-workflow-run`) sit directly under `.github/actions`, so the glob covers them.

### DIFF
--- a/.github/actions/setup_python_env/action.yml
+++ b/.github/actions/setup_python_env/action.yml
@@ -45,7 +45,7 @@ runs:
         enable-cache: true
 
     - name: Setup mise
-      uses: jdx/mise-action@c37c93293d6b742fc901e1406b8f764f6fb19dac # v2.4.4
+      uses: jdx/mise-action@7e36c90d9ab29c415a2384db3006f3ec8a8cc654 # v4.2.4
       with:
         install: false
 
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,7 +4,11 @@ updates:
     - package-ecosystem: github-actions
       cooldown:
         default-days: 14
-      directory: /
+      # "/" scans .github/workflows; the globs cover composite actions under
+      # .github/actions, which the "/" root does not reach.
+      directories:
+        - /
+        - /.github/actions/*
       schedule:
           interval: monthly
     - package-ecosystem: pip


pgoslatara:
## 🔍 AI PR Review
> **Extends Dependabot scanning to composite actions under `.github/actions/` and bumps `jdx/mise-action` to v4.2.4 (Node.js 24 support).**

**Type:** `chore`

### Summary
Dependabot's `directory: /` only scans `.github/workflows/`, so pinned actions inside composite actions under `.github/actions/` were invisible to automated updates. This PR switches to `directories` (plural) with a glob pattern to cover those subdirectories. It also bumps the one composite-action dependency that was already stale: `jdx/mise-action` from v2.4.4 to v4.2.4.

### File Changes
| `.github/dependabot.yml` | Modified | Switch `directory` to `directories` list, adding `/.github/actions/*` glob so Dependabot scans composite actions |
| `.github/actions/setup_python_env/action.yml` | Modified | Bump `jdx/mise-action` from `v2.4.4` (SHA `c37c93...`) to `v4.2.4` (SHA `7e36c9...`) |

### Review Findings
- ✅ No issues with **cross-file version consistency** — the old SHA/version (`c37c93...` / `v2.4.4`) appears only in the updated file; no stale references elsewhere.
- ✅ No issues with **security** — action is pinned to a full commit SHA with a version comment, following best practice.
- ✅ No issues with **style** — the `dependabot.yml` comment clearly explains *why* the glob is needed; indentation matches the existing file.
- ✅ No issues with **breaking changes** — `mise-action` is used with `install: false` (cache-only), so the major-version jump (v2 → v4) has no functional impact on the setup step.
- ✅ No issues found. Reviewed for Python quality, check structure, tests, security, style, breaking changes, and documentation (no Python or test files changed in this PR).
---

PR #1066, opened 2026-08-18
fix(catalog): case-insensitive column matching for Snowflake

### LINKED ISSUE #1064: [Bug] Snowflake source catalog checks mis-handle wrapped/uppercased columns
### Current Behavior

Source catalog checks can flag documented Snowflake source columns as undocumented when catalog metadata is wrapped or when catalog column names are uppercased.

### Expected Behavior

Source column documentation checks should accept wrapped catalog source objects and compare column names case-insensitively.

### Steps To Reproduce

1. Run dbt-bouncer against a Snowflake project.
2. Use catalog output where source columns are uppercased or exposed through a nested wrapper.
3. Observe false undocumented-column warnings from `check_source_columns_are_all_documented`.

### Relevant log output

```shell
check_source_columns_are_all_documented: ... WARN
```

### Environment

```markdown
- OS: Windows
- Python: 3.12.5
- dbt-bouncer: v3.8.0
```

### Add your company's website here to be added to our list of [users](https://godatadriven.github.io/dbt-bouncer/stable/users/#users).

_No response_

pgoslatara:
Several catalog checks compared catalog column names against the properties file case-sensitively. On Snowflake the catalog stores column names in uppercase while the `.yml` properties file uses lowercase, so documented columns were reported as undocumented.

This adds a `case_sensitive` parameter to every affected catalog check. It defaults to `false` for `dbt-snowflake` and `true` otherwise, following the pattern already used by `check_columns_are_all_documented` (CA006). The four catalog column checks now behave consistently:

- `check_seed_columns_are_all_documented` (CA001)
- `check_source_columns_are_all_documented` (CA004)
- `check_columns_are_documented_in_public_models` (CA007)

`schema.json` is regenerated to expose the new parameter. Unit tests cover, for each check, the uppercase Snowflake match, a genuinely undocumented column that must still fail, and an explicit opt-in on a non-Snowflake adapter.

Resolves #1064

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -777,7 +777,7 @@
     },
     "CheckColumnsAreDocumentedInPublicModels": {
       "additionalProperties": false,
-      "description": "Columns should have a populated description in public models.\n\n!!! info \"Rationale\"\n\n    Public models form the stable, documented API of a dbt project \u2014 they are the models that external consumers, BI tools, and other projects are expected to query directly. Undocumented columns in a public model undermine this contract: consumers cannot tell what a field means, which ID to join on, or whether a flag is nullable. Requiring column descriptions on public models ensures the published interface is self-explanatory and trustworthy, which is especially important as teams scale and data catalogues surface model metadata automatically.\n\nReceives:\n    catalog_node (CatalogNodeEntry): The CatalogNodeEntry object to check.\n    min_description_length (int | None): Minimum length required for the description to be considered populated.\n    models (list[ModelNode]): List of ModelNode objects parsed from `manifest.json`.\n\nOther Parameters:\n    description (str | None): Description of what the check does and why it is implemented.\n    exclude (str | list[str] | None): Regex pattern(s) to match the model path. Model paths that match any pattern will not be checked.\n    include (str | list[str] | None): Regex pattern(s) to match the model path. Only model paths that match any pattern will be checked.\n    severity (Literal[\"error\", \"warn\"] | None): Severity level of the check. Default: `error`.\n\nExample(s):\n    ```yaml\n    catalog_checks:\n        - name: check_columns_are_documented_in_public_models\n    ```",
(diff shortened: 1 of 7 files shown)


pgoslatara:
## 🔍 AI PR Review
> **Fixes false-positive column-documentation failures on Snowflake by adding case-insensitive column matching across three catalog checks.**

**Type:** `bugfix`

### Summary
Snowflake stores column names as uppercase in `catalog.json`, while `manifest.json` (the properties file) typically has lowercase names. This case mismatch caused catalog checks to wrongly report documented columns as undocumented. This PR adds a `case_sensitive` parameter (default `True`) to three catalog checks and auto-detects the Snowflake adapter to disable case-sensitivity automatically. The same case-insensitive path can be opted into explicitly for other adapters via `case_sensitive: false` in the config.

All three check files follow the same pattern: detect Snowflake → force `case_sensitive = False` → lowercase both sides before comparing. Tests cover the Snowflake auto-detect path, explicit opt-in on non-Snowflake adapters, and ensure genuinely undocumented columns still fail.

<details>
<summary>📄 File Changes (7 files)</summary>

| `src/dbt_bouncer/checks/catalog/check_catalog_seeds.py` | Modified | Added `case_sensitive` param with Snowflake auto-detection for seed column matching |
| `src/dbt_bouncer/checks/catalog/check_catalog_sources.py` | Modified | Added `case_sensitive` param with Snowflake auto-detection for source column matching |
| `src/dbt_bouncer/checks/catalog/columns/description.py` | Modified | Added `case_sensitive` param with Snowflake auto-detection for public model column description check |
| `tests/unit/checks/catalog/test_catalog_seeds.py` | Modified | Added Snowflake-specific tests: auto-detect pass, undocumented still fails, explicit opt-in |
| `tests/unit/checks/catalog/test_catalog_sources.py` | Modified | Added case-mismatch failure test + Snowflake-specific test class with 3 scenarios |
| `tests/unit/checks/catalog/columns/test_description.py` | Modified | Added Snowflake-specific tests: auto-detect pass, undocumented still fails, explicit opt-in |
| `schema.json` | Modified | Added `case_sensitive` boolean property to three check schemas, updated descriptions |

</details>

### Review Findings

- 🟡 **Suggestion:** `check_catalog_seeds.py:83`, `check_catalog_sources.py:38`, `columns/description.py:184` — The Snowflake auto-detection unconditionally overrides the user-supplied `case_sensitive` value. Since the parameter default is `True` and the override always sets `False` on Snowflake, there is no way for a Snowflake user to opt *into* case-sensitive matching. This is unlikely to be needed in practice (Snowflake always uppercases catalog columns), but it means `case_sensitive: true` in a Snowflake config is silently ignored. Consider documenting this explicitly or only overriding when the user hasn't set the parameter.

- 🟡 **Suggestion:** `check_catalog_seeds.py:83`, `check_catalog_sources.py:38`, `columns/description.py:184` — The Snowflake detection logic (`if ctx.manifest_obj.manifest.metadata.adapter_type in ["snowflake"]`) is duplicated across three checks. If more adapters need case-insensitive matching in the future (e.g. a hypothetical uppercase adapter), a shared helper like `_is_case_insensitive_adapter(ctx)` would reduce maintenance burden. Not blocking, but worth considering.

- 🟡 **Suggestion:** `test_catalog_sources.py:97-125` (the `case_mismatch` parametrized test) — This new case tests that uppercase catalog columns *fail* on non-Snowflake adapters (case-sensitive mode), which validates the default behavior well. However, the test name `case_mismatch` is generic — consider a more descriptive ID like `case_mismatch_fails_without_snowflake` to clarify intent when reading test output.

- ✅ No issues with **Python quality** — type hints on new parameter, clean imports, proper use of keyword-only arguments.
- ✅ No issues with **check structure** — `@check` decorator used correctly, `fail()` called properly, `case_sensitive` is a keyword-only param as expected.
- ✅ No issues with **test coverage** — all three checks have happy path (Snowflake auto-detect), unhappy path (genuinely undocumented column still fails on Snowflake), and explicit opt-in on non-Snowflake.
- ✅ No issues with **security** — no secrets, no unsafe deserialization, no injection vectors.
- ✅ No issues with **style** — consistent snake_case, parameter ordering follows convention.
- ✅ No issues with **breaking changes** — `case_sensitive` defaults to `True`, preserving existing behavior for non-Snowflake users. Schema additions are additive.
- ✅ No issues with **documentation** — docstrings updated with new `Parameters` section, schema.json regenerated.
---

michaelconan:
Go raibh maith agat @pgoslatara 🍻

PR #1090, opened 2026-08-28
feat: read `run` options from environment variables

### LINKED ISSUE #1089: [Feature] Allow `run` options to be set via environment variables
### Problem

`dbt-bouncer` can only be told to write machine-readable results through the command line: `--output-file`, `--output-format`, `--output-only-failures`. There is no config-file equivalent (`output_file` is not a `dbt-bouncer.yml` key, and `Reporter` writes structured output only when `output_file` is set), and no environment variable.

That is a problem for anything that runs `dbt-bouncer` through the pre-commit hook rather than directly, because **`pre-commit run` has no argument passthrough** — the only argv it will use is the `args:` list in the project's own `.pre-commit-config.yaml`, and `pre-commit run --help` offers nothing but `-c/--config`.

Concretely: I maintain a dbt code-review tool that reads the project's manifest and turns each `dbt-bouncer` failure into a review finding. It has to run the exact version CI runs, so it resolves the hook's pinned `repo@rev` out of pre-commit's own repo cache — but then it has to execute that binary directly, bypassing `pre-commit run`, purely to be able to pass `--output-file`. The alternatives are both worse:

- Ask projects to put `--output-file /tmp/...` in their committed `.pre-commit-config.yaml`, which changes what CI runs for everyone.
- Synthesize a throwaway pre-commit config that copies the project's pinned `repo`, `rev` and `additional_dependencies` and appends the output arguments, then run `pre-commit run --config <that>`. This works, but it means parsing and faithfully re-emitting someone else's hook block.

Any wrapper that cannot control argv — pre-commit, and I would guess some CI images and task runners too — hits the same wall.

### Proposal

Read the `run` options from environment variables, using typer's built-in `envvar=`. There is precedent in the project already: `DBT_BOUNCER_CONFIG_FILE`, `DBT_PROJECT_DIR`, `DBT_BOUNCER_DISABLE_CONF_CACHE` and `DBT_BOUNCER_BENCH_MODELS` are all read from the environment today.

| Option | Proposed variable |
| --- | --- |
| `--check` | `DBT_BOUNCER_CHECK` |
| `--only` | `DBT_BOUNCER_ONLY` |
| `--dry-run` | `DBT_BOUNCER_DRY_RUN` |
| `--output-file` | `DBT_BOUNCER_OUTPUT_FILE` |
| `--output-format` | `DBT_BOUNCER_OUTPUT_FORMAT` |
| `--output-only-failures` | `DBT_BOUNCER_OUTPUT_ONLY_FAILURES` |
| `--show-all-failures` | `DBT_BOUNCER_SHOW_ALL_FAILURES` |
| `-v` / `--verbosity` | `DBT_BOUNCER_VERBOSITY` |
| `--create-pr-comment-file` | `DBT_BOUNCER_CREATE_PR_COMMENT_FILE` |

Command-line arguments keep precedence over the environment, which is click's default resolution order, so nothing changes for anyone who does not set the variables.

This would make the following work:

```bash
DBT_BOUNCER_OUTPUT_FILE=results.json \
DBT_BOUNCER_OUTPUT_FORMAT=json \
DBT_BOUNCER_OUTPUT_ONLY_FAILURES=true \
DBT_BOUNCER_ONLY=manifest_checks \
  pre-commit run dbt-bouncer --all-files
```

### Two deliberate exclusions, which I would like your opinion on

**`--config-file`.** It is already environment-configurable via `DBT_BOUNCER_CONFIG_FILE`, but through the bespoke logic in `get_config_file_path()`, whose documented precedence puts the environment variable *below* the CLI flag. Adding `envvar=` to the typer option would change that behaviour rather than duplicate it: `detect_config_file_source()` decides `ConfigFileSource.COMMANDLINE` by comparing the value against the three default filenames, and never consults click's parameter source, so an environment-provided path would be reclassified as `COMMANDLINE`. It would jump from precedence step 2 to step 1 and start raising `DbtBouncerConfigError` when the path does not exist. So I would leave that option exactly as it is and just document it alongside the new ones.

**The other subcommands.** `--output-format` also exists on `list` and `explain`, but with a different enum (`text|json` rather than `csv|json|junit|sarif|tap`). A single shared `DBT_BOUNCER_OUTPUT_FORMAT` would therefore break whichever subcommand the value does not suit. Options I can see: scope the variables to `run` only (my preference, and what the table above assumes), or namespace the others as `DBT_BOUNCER_LIST_OUTPUT_FORMAT` and so on. Happy to go either way.

### Willing to contribute

Yes — the change is written and green against `main`: `envvar=` on the nine options on `run` plus the identical set on the legacy no-subcommand callback, with `docs/cli.md` and `docs/configuration.md` updated and eight new unit tests in `tests/unit/test_main.py` (seven of which fail without the change). `prek run --all-files`, `pytest tests/unit` (1752 passed) and the Zensical docs build are all clean. I will open the PR referencing this issue.

I am raising the issue first because environment variable names are a public API surface, so it seemed worth giving you a chance to veto the naming — and to pick between the two scoping options above — before the code lands.

ms32035:
Closes #1089.

### What

Adds typer's `envvar=` to every option on `dbt-bouncer run`, and to the identical set on the legacy no-subcommand callback, so each can also be set with a `DBT_BOUNCER_*` environment variable.

| Option | Environment variable |
| --- | --- |
| `--check` | `DBT_BOUNCER_CHECK` |
| `--only` | `DBT_BOUNCER_ONLY` |
| `--dry-run` | `DBT_BOUNCER_DRY_RUN` |
| `--output-file` | `DBT_BOUNCER_OUTPUT_FILE` |
| `--output-format` | `DBT_BOUNCER_OUTPUT_FORMAT` |
| `--output-only-failures` | `DBT_BOUNCER_OUTPUT_ONLY_FAILURES` |
| `--show-all-failures` | `DBT_BOUNCER_SHOW_ALL_FAILURES` |
| `-v` / `--verbosity` | `DBT_BOUNCER_VERBOSITY` |
| `--create-pr-comment-file` | `DBT_BOUNCER_CREATE_PR_COMMENT_FILE` |

### Why

`pre-commit run` has no argument passthrough — the only argv a hook ever sees is the `args:` list committed in the project's `.pre-commit-config.yaml`. So a tool that invokes dbt-bouncer through its pre-commit hook cannot ask for `--output-file`/`--output-format`, and there is no other way to reach structured output (`output_file` is not a config file key, and `Reporter` writes an output file only when it is set).

I hit this building a dbt review tool that turns each dbt-bouncer failure into a review finding. It has to run the version CI runs, so it resolves the hook's pinned `repo@rev` from pre-commit's own repo cache — but then it must execute that binary directly, going around `pre-commit run`, purely to pass `--output-file`. With this change that becomes:

```bash
DBT_BOUNCER_OUTPUT_FILE=results.json \
DBT_BOUNCER_OUTPUT_FORMAT=json \
DBT_BOUNCER_OUTPUT_ONLY_FAILURES=true \
DBT_BOUNCER_ONLY=manifest_checks \
  pre-commit run dbt-bouncer --all-files
```

There is precedent for environment-driven configuration in the project already: `DBT_BOUNCER_CONFIG_FILE`, `DBT_PROJECT_DIR`, `DBT_BOUNCER_DISABLE_CONF_CACHE` and `DBT_BOUNCER_BENCH_MODELS`.

### Backwards compatibility

None intended. Command-line arguments keep precedence over the environment, which is click's normal resolution order, so behaviour is unchanged for anyone who does not set the variables. The only visible difference otherwise is an `[env var: …]` line in `--help`.

### Two deliberate exclusions

**`--config-file` is left as it is.** It already reads `DBT_BOUNCER_CONFIG_FILE`, but through `get_config_file_path()`, whose documented precedence puts the variable *below* the CLI flag. Adding `envvar=` would change that behaviour rather than duplicate it: `detect_config_file_source()` decides `ConfigFileSource.COMMANDLINE` by comparing the value against the three default filenames and never consults click's parameter source, so an environment-provided path would be reclassified as `COMMANDLINE` — jumping from precedence step 2 to step 1, and newly raising `DbtBouncerConfigError` when the path does not exist. It is documented alongside the new variables instead, with a pointer to the config file page.

**`validate`, `list` and `explain` are not covered.** `--output-format` exists on `list` and `explain` too, but with a different enum (`text|json` rather than `csv|json|junit|sarif|tap`), so one shared `DBT_BOUNCER_OUTPUT_FORMAT` would break whichever subcommand the value did not suit. Happy to follow up with namespaced variables (`DBT_BOUNCER_LIST_OUTPUT_FORMAT`, …) if you would prefer full coverage.

### Tests

Eight new tests in `tests/unit/test_main.py`, following the existing `test_cli_output_formats` structure but passing values through `CliRunner.invoke(..., env=…)` instead of argv:

- output options read from the environment, parametrized over both `dbt-bouncer run` and the legacy invocation;
- `--check` read from the environment, both invocations;
- `--dry-run` as a boolean flag from the environment;
- `--verbosity` through the `count=True` counter;
- an explicit CLI flag beating the same variable;
- an invalid enum value failing as a usage error that names the variable.

Seven of the eight fail with the `src/` changes reverted, so they discriminate.

### Checks run locally

- `uv run pytest tests/unit` — 1752 passed
- `uv run pytest tests/integration` — 21 passed
- `uv run prek run --all-files` — all hooks pass (ruff, ruff-format, ty, cspell, rumdl, bandit, schema-check, …)
- `uv run zensical build` — "No issues found"

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -21,6 +21,7 @@ All the main CLI options (`--check`, `--only`, `--output-file`, etc.) work with 
 **Type:** Path
 **Default:** `dbt-bouncer.yml`
 **Required:** No
+**Environment variable:** `DBT_BOUNCER_CONFIG_FILE` (see [Config file](configuration.md))
 
 Specifies the location of the YAML configuration file containing your dbt-bouncer checks.
 
@@ -35,6 +36,7 @@ dbt-bouncer run --config-file config/checks.yml
 **Type:** Flag
 **Default:** False
 **Required:** No
+**Environment variable:** `DBT_BOUNCER_DRY_RUN`
 
 When passed, assembles the full check list as normal but prints a summary table showing the check name, resource type, and count for each check that would run — then exits with code 0 without executing any checks. Useful for previewing which checks are in scope before a full run.
 
@@ -62,6 +64,7 @@ Dry run complete. 2524 check(s) would run.
 **Type:** String (comma-separated)
 **Default:** Empty (runs all checks)
 **Required:** No
+**Environment variable:** `DBT_BOUNCER_CHECK`
 
 Limits the checks run to specific check names. Multiple checks can be specified as a comma-separated list.
 
@@ -80,6 +83,7 @@ dbt-bouncer run --check check_model_names,check_source_freshness_populated
 **Type:** String (comma-separated)
 **Default:** Empty (runs all categories)
 **Required:** No
+**Environment variable:** `DBT_BOUNCER_ONLY`
 
 Limits the checks run to specific categories. Multiple categories can be specified as a comma-separated list.
 
@@ -98,6 +102,7 @@ dbt-bouncer run --only catalog_checks,manifest_checks
 **Type:** Path
 **Default:** None (no output file is written)
 **Required:** No
+**Environment variable:** `DBT_BOUNCER_OUTPUT_FILE`
 
 Specifies the location where check metadata will be saved. If not provided, no structured output file is written.
 
@@ -113,6 +118,7 @@ dbt-bouncer run --output-file results/check-results.json
 **Options:** `csv`, `json`, `junit`, `sarif`, `tap`
 **Default:** `json`
 **Required:** No
+**Environment variable:** `DBT_BOUNCER_OUTPUT_FORMAT`
 
 Specifies the format for the output file. Requires `--output-file` to be set.
 
@@ -140,6 +146,7 @@ dbt-bouncer run --output-format sarif --output-file results.sarif
 **Type:** Flag
 **Default:** False
 **Required:** No
+**Environment variable:** `DBT_BOUNCER_OUTPUT_ONLY_FAILURES`
 
 When passed, only failures will be included in the output file. Successful checks are omitted.
 
@@ -154,6 +161,7 @@ dbt-bouncer run --output-file results.json --output-only-failures
 **Type:** Flag
 **Default:** False
 **Required:** No
+**Environment variable:** `DBT_BOUNCER_SHOW_ALL_FAILURES`
 
 When passed, all failures will be printed to the console, even if an output file is specified.
 
@@ -168,6 +176,7 @@ dbt-bouncer run --show-all-failures
 **Type:** Counter
 **Default:** 0
 **Required:** No
+**Environment variable:** `DBT_BOUNCER_VERBOSITY`
 
 Controls the verbosity of logging output. Can be specified multiple times to increase verbosity.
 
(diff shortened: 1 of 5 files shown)


pgoslatara (approved):
Thanks for the PR @ms32035! Agree with your decision making here so no need for changes. The next release of `dbt-bouncer` will be a major release so including these breaking changes is good timing. Note that this release will occur when `dbt-core` 2.0 is released, hopefully that happens soon.

Issue #1110, opened 2026-09-14
tests: further improve integration tests


mattcristal:
## Premise

With AI-driven development, integrations are very important to guarantee that the application still works and works in the expected way.

In some situations, it is expected that some unit tests can break / might be re-written by AI(?)

## TODO

Further enhance integration tests for dbt-bouncer:

https://github.com/godatadriven/dbt-bouncer/tree/main/tests/integration

mattcristal:
I will also provide some ideas or suggest some PRs

mattcristal:
@pgoslatara 

https://github.com/godatadriven/dbt-bouncer/pull/1111

See this one.

mattcristal:
@pgoslatara for now, I am happy with the progress and I will closed the issue

PR #1111, opened 2026-09-14
tests: add integration test for cli exit codes

mattcristal:
### Intro

Severity / exit codes. No integration test asserts that a severity: warn failure yields exit 0 while error yields 1, or that the WARN=n summary line is right. 

See also: 

https://github.com/godatadriven/dbt-bouncer/issues/1110


mattcristal:
@pgoslatara 

I am not very satisfied with this PR, actually.

Converted back to draft, but you see where I am going.

PR #1117, opened 2026-09-16
tests: add integration test for cli exit codes

mattcristal:
## What was done

tests: add integration test for cli exit codes


mattcristal:
@pgoslatara now you can review

pgoslatara (approved):

